PR #1, opened 2021-08-20
Allow pod annotations for the daemonset

treydock:
This is useful when using Prometheus but not with an operator so can tell Prometheus to scrape the pod port.


nikkon-dev (approved):

treydock:
@nikkon-dev Is it possible to get the Helm chart version bumped and pushed to Helm chart repo for this pull request and #2?  I wasn't sure if there was a predetermined release cadence or if this could be done now.

PR #2, opened 2021-08-21
Allow disabling service

treydock:
We scrape the pod port directly. Plus scraping the service doesn't really work with daemonset since the service will point to only one of the daemonset pods, and when more than one node exists with GPUs, this ends up not scraping more than one node.

### DIFF
--- a/deployment/templates/service.yaml
+++ b/deployment/templates/service.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.service.enable }}
 # Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -33,3 +34,4 @@ spec:
     protocol: TCP
   selector:
     {{- include "dcgm-exporter.selectorLabels" . | nindent 4 }}
+{{- end }}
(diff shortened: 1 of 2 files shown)


nikkon-dev:
How would Helm like an empty template? For example, `helm template` needs `--output-dir` argument to ignore empty YAML files.

treydock:
Helm does just fine, I did this and get correctly rendered deployment:

```
$ helm template dcgm-exporter deployment --set service.enable=false
```

This rendered my local changes without the Service.  This pattern of not deploying a template is already being used in this Helm chart:

https://github.com/NVIDIA/dcgm-exporter/blob/50ccb1a609153bed097ccfb849797a5ab8f7b23b/deployment/templates/service-monitor.yaml#L1

My local Helm deployment of this exporter sets that serviceMonitor to not be included and it works great.  Helm doesn't care if one file is empty, since it renders all the files together so one empty file is going to be like that file doesn't exist.

treydock:
I moved the boolean for enable/disable service to same place (above comments) like the serviceMonitor file to make things consistent.

nikkon-dev:
@treydock, Could you sign-off you commit, please?

treydock:
Updated commit to be signed off.

PR #10, opened 2021-09-08
Enable ConfigMap support with "-m <namespace>:<name>"

glowkey:
Include default configmap as part of helm chart.
    Update example values to explain configmap usage.
    Include tests for parsing configmap.

Signed-off-by: Douglas Wightman <dwightman@nvidia.com>


pkg/dcgmexporter/parser.go
```diff
@@ -127,3 +150,40 @@ func recordIsCommentOrEmpty(s []string) bool {
...
+		return nil, fmt.Errorf("Malformed configmap-data: %s", c.ConfigMapData)
+	}
+
+	var cm *corev1.ConfigMap
+	cm, err := kubeClient.CoreV1().ConfigMaps(parts[0]).Get(context.TODO(), parts[1], metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("Could not retrieve ConfigMap '%s': %v", c.ConfigMapData, err)
+	}
+
+	r := csv.NewReader(strings.NewReader(cm.Data["metrics"]))
```
nikkon-dev, line 166: I'd suggest to validate that "metrics" are actually present in the map:
if metrics, ok := cm.Data["metrics"]; ok {
...
} else {
Report an error here
}

pkg/dcgmexporter/parser.go
```diff
@@ -17,23 +17,46 @@
...
+	}
+
 	if err != nil {
-		fmt.Printf("Error: %v\n", err)
-		return nil, err
+		logrus.Infof("%v, falling back to metric file %s", err, c.CollectorsFile)
+
+		records, err = ReadCSVFile(c.CollectorsFile)
+		if err != nil {
+			fmt.Printf("Error: %v\n", err)
```
nikkon-dev, line 54: Shouldn't we log this to the logrus as well instead?

pkg/dcgmexporter/parser.go
glowkey: I've added logic and an error if 'metrics' is not present.

cmd/dcgm-exporter/main.go
```diff
@@ -53,6 +54,7 @@ var (
 	CLIDevices             = "devices"
 	CLINoHostname          = "no-hostname"
 	CLIUseFakeGpus         = "fake-gpus"
+	ClIConfigMapData       = "configmap-data"
```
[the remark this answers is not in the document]
glowkey: Good catch. Fixed.

PR #18, opened 2021-10-01
DCGM-2405. Report on GPUs and GPU instances when there are MIG and no…

dbeer:
…n-MIG GPUs.


glowkey:
I believe the CLI usage help should be updated to reflect the new behavior of "-d f", ideally using the term "migStrategy" to associate its usage with the rest of the k8s stack.

glowkey (dismissed):

glowkey (approved):

PR #20, opened 2021-10-07
Make ConfigMap values dynamic and use as default metric list

hfurkanvural:
Signed-off-by: H. Furkan Vural <hfurkanvural@gmail.com>

### DIFF
--- a/deployment/templates/daemonset.yaml
+++ b/deployment/templates/daemonset.yaml
@@ -62,13 +62,13 @@ spec:
       - name: "pod-gpu-resources"
         hostPath:
           path: "/var/lib/kubelet/pod-resources"
+      - name: "exporter-metrics"
+        configMap:
+          name: "exporter-metrics-config-map"
       {{- range .Values.extraHostVolumes }}
       - name: {{ .name | quote }}
         hostPath:
           path: {{ .hostPath | quote }}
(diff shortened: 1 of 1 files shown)


deployment/templates/metrics-configmap.yaml
```diff
@@ -3,7 +3,10 @@ kind: ConfigMap
 metadata:
   name: exporter-metrics-config-map
 data:
-  metrics: |
```
glowkey, line 6: Changing this line will also require changing this line in the parser.go file: https://github.com/NVIDIA/dcgm-exporter/blob/main/pkg/dcgmexporter/parser.go#L166

deployment/templates/daemonset.yaml
```diff
@@ -66,6 +66,9 @@ spec:
       - name: {{ .name | quote }}
         hostPath:
           path: {{ .hostPath | quote }}
+      - name: "exporter-metrics"
+        configMap:
+          name: "exporter-metrics-config-map"
```
glowkey, line 71: These lines are inside the range and will cause an error when installing the helm chart. They should be moved above or below.

hfurkanvural: Thanks for heads up! Fixed.

PR #60, opened 2022-04-07
makefile improvements

lukeyeager:
NOTE: this makes two breaking changes to the 'install' target:
1. Files are now installed to "standard" locations such as /usr/local/bin/ instead of /usr/bin/. To get back to the old behavior:
    `sudo make prefix=/usr sysconfdir=/etc install`.
1. The install target no longer depends on the binary target, which could break build workflows which only run `make install`. This could be fixed with additional work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,6 @@ SYSCONFDIR := $(abspath $(DESTDIR))$(sysconfdir)
 GO      ?= go
 GOFLAGS ?=
 
-MKDIR    ?= mkdir
 REGISTRY ?= nvidia
 
 DCGM_VERSION   := 2.3.5
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+prefix     ?= /usr/local
+bindir     ?= $(prefix)/bin
+sysconfdir ?= $(prefix)/etc
+
+BINDIR     := $(abspath $(DESTDIR))$(bindir)
+SYSCONFDIR := $(abspath $(DESTDIR))$(sysconfdir)
+
 MKDIR    ?= mkdir
 REGISTRY ?= nvidia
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -33,15 +42,14 @@ MAIN_TEST_FILES := pkg/dcgmexporter/system_info_test.go
...
-	go test ./...
+	$(GO) test ./...
 
-install: binary
-	install -m 557 cmd/dcgm-exporter/dcgm-exporter /usr/bin/dcgm-exporter
-	install -m 557 -D ./etc/default-counters.csv /etc/dcgm-exporter/default-counters.csv
-	install -m 557 -D ./etc/dcp-metrics-included.csv /etc/dcgm-exporter/dcp-metrics-included.csv
+install:
+	install -m 0755 -D -t $(BINDIR) cmd/dcgm-exporter/dcgm-exporter
+	install -m 0644 -D -t $(SYSCONFDIR) etc/default-counters.csv etc/dcp-metrics-included.csv
```
nikkon-dev, line 52: dcgm-exporter by default will look for the default-counters.csv in the /etc/dcgm-exporter/ location:
https://github.com/NVIDIA/dcgm-exporter/blob/0869a2ac03c73e82715438f44d2ecf2491049139/cmd/dcgm-exporter/main.go#L96

Makefile
```diff
@@ -12,7 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-MKDIR    ?= mkdir
+prefix     ?= /usr/local
+bindir     ?= $(prefix)/bin
+sysconfdir ?= $(prefix)/etc
```
nikkon-dev, line 17: Is this standard? I do not see it in the FHS list.

Makefile
lukeyeager: You're right - that's an error on my part. Will fix.

Makefile
lukeyeager: Honestly I just copied it from Jon who usually tracks these conventions closely: https://github.com/NVIDIA/enroot/blob/v3.4.0/Makefile#L5

But yes, in this case it does look like it's standard:
> ‘sysconfdir’
> The directory for installing read-only data files that pertain to a single machine–that is to say, files for configuring a host. ...
> https://www.gnu.org/software/make/manual/html_node/Directory-Variables.html

lukeyeager: I suppose since we're looking at that file I should point out that files like this should really be under `/usr/share/` rather than `/etc/` (see the definition of 'datarootdir'), but I don't personally care to make that breaking change today.

nikkon-dev: The problem with that breaking change is that there are many customers who mount default-counters.csv into /etc/dcgm-exporter/ location via docker -v command-line argument. Such change will break their use cases as they will have to either change where the file is mounted inside the container or change the dcgm-exporter command-line arguments. I would consider both scenarios as a breaking change that needs a major version release at least.

lukeyeager: Right, yes, I agree. But we're choosing not to make the breaking change, right? I was just making a pedantic point. I'm not arguing the situation change that now or even ever.

nvvfedorov:
The PR is stale because there hasn't been any activity for the last two years.

PR #88, opened 2022-07-29
Fix the type of the PCIE_TX/RX metrics and provide more accurate description.

nikkon-dev:
DCGM_FI_PROF_PCIE_TX_BYTES and DCGM_FI_PROF_PCIE_RX_BYTES are PCIe bandwidths that are computed for each scraping interval. The types of those metrics were mistakenly specified as 'counter' instead of 'gauge.' 
The descriptions of those metrics were also updated according to the official DCGM documentation.The descriptions of those metrics were also updated according to the officidal DCGM documentation.


glowkey:
Should this change also be made to "etc/default-counters.csv"?

nikkon-dev:
> Should this change also be made to "etc/default-counters.csv"?

I do not see them in the default-counters.csv. These are DCP metrics

glowkey:
You're right. Confused it with DCGM_FI_DEV_PCIE_TX and _RX.

glowkey (approved):

Issue #92, opened 2022-08-18
Pod labels are not propagated for MIGs on GKE [Flag "..._GPU_ID_TYPE" has no effect for MIG devices]


suffiank:
MIG device metrics on GKE are missing namespace, pod, and container labels.
This flag does not appear to have an effect for MIG devices:
```
- name: "DCGM_EXPORTER_KUBERNETES_GPU_ID_TYPE"
              value: "device-name"
```

The GKE device plug-in exports MIG devices with a name scheme like "nvidia0/gi7" "nvidia0/gi8" (reference [here](https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/50c6afb173f99fcb20ad62f14b175dbdc1a79a2d/pkg/gpu/nvidia/mig/mig.go#L193)).
While DCGM exporter seems looking for "MIG-GPU-9456f8ea-1442-2159-bdc4-d87b9893c817/12/0" (reference [here](https://github.com/NVIDIA/dcgm-exporter/blob/bdcfaecbcfba2e3a48bf96231edcbd91d109f000/pkg/dcgmexporter/kubernetes.go#L155))

Can we extend the flag to handle the MIG case as well? The flag is applied [here](https://github.com/NVIDIA/dcgm-exporter/blob/bdcfaecbcfba2e3a48bf96231edcbd91d109f000/pkg/dcgmexporter/types.go#L129).

suffiank:
Hi @nikkon-dev 

Are you sure you got that right about the purpose of the flag? (reference [here](https://github.com/NVIDIA/dcgm-exporter/blob/574d63d717af9a4447092070d9501ec1eb83873e/cmd/dcgm-exporter/main.go#L144)) So far as I can tell this is its only function in the code base.

It takes less code to extend the application of the flag to assume the "device-name" matches "nvidia0/gi7" etc. for MIGs. But in any case I guess scanning for the GKE format is better. I opened https://github.com/NVIDIA/dcgm-exporter/pull/97. 

Take a look? Thanks.

nikkon-dev:
> Are you sure you got that right about the purpose of the flag? (reference [here](https://github.com/NVIDIA/dcgm-exporter/blob/574d63d717af9a4447092070d9501ec1eb83873e/cmd/dcgm-exporter/main.go#L144)). So far as I can tell this is its only function in the code base.

[...]

PR #97, opened 2022-08-24
Fix propagation of pod labels on GKE with MIG devices by scanning for GKE device ID format

suffiank:


elfinhe:
#92

suffiank:
Here's an alternative fix that simply assumes DCGM_EXPORTER_KUBERNETES_GPU_ID_TYPE == "device-name" in case of MIGs will have device IDs match the GKE form "nvidia0/gi7".
https://github.com/NVIDIA/dcgm-exporter/compare/main...suffiank:dcgm-exporter:sufkha/migs-on-gke

I'm not sure how many other device plugins are out there.

pkg/dcgmexporter/kubernetes.go
```diff
@@ -35,6 +36,8 @@ var (
 	socketPath = socketDir + "/kubelet.sock"
 
 	connectionTimeout = 10 * time.Second
+
+	gkeMigDeviceIdRegex = regexp.MustCompile(`^nvidia(?P<gpu>[0-9]+)/gi(?P<gi>[0-9]+)$`)
```
nikkon-dev, line 40: You do not use the group names in the code below, but those make the regex visually more difficult to read. Could we remove the group names or use them in the code below?

suffiank: Yeah, makes sense - removed.

nikkon-dev (approved):
LGTM

Issue #78, opened 2022-07-08
Add Kubernetes node name to exported labels


neggert:
Right now, when dcgm-exporter is deployed in Kubernetes (we're using gpu-operator), the `Hostname` label is set to the pod name, which is not particularly useful. I'd like to suggest either:

* Adding a new label `node` or
* Using different logic to populate `Hostname` when running in Kubernetes

It should be fairly straightforward to inject the node name into the container using the Downward API.

alex-g-tejada:
Would it be possible to have labels in general added as well? We have kubernetes pods that have label.app=dev for example and these are not visible in DCGM metrics

k0nstantinv:
@neggert as workaround right now you can do it from the Prometheus dcgm-exporter job side. You need kube-state-metrics exporter or something like that, then you can do relabel_config. For example:
```
- job_name: dcgm-exporter
  scrape_interval: 30s
  scrape_timeout: 10s
  kubernetes_sd_configs:
    - role: endpoints
  relabel_configs:
    - source_labels: [__meta_kubernetes_endpoints_name]
      regex: 'dcgm-exporter'
      action: keep
    - source_labels: [__meta_kubernetes_namespace]
      action: replace
      target_label: namespace
    - source_labels: [__meta_kubernetes_pod_name]
      action: replace
      target_label: pod_name
    - source_labels: [__meta_kubernetes_pod_node_name]
      action: replace
      target_label: node
```
result
```
DCGM_FI_DEV_FB_FREE{UUID="GPU-16e319ba-0b7d-3a4b-e35f-915bf484870f", cluster="test-cluster01", device="nvidia0", gpu="0", instance="10.109.133.139:9400", job="dcgm-exporter", namespace="dcgm-exporter", node="node06", pod_name="dcgm-exporter-k22f5"}
```

PlayMTL:
Hey, i have solved this by adding some relabeling to my ServiceMonitor.

I also overwrote the instance label so I don't have to customize my grafana boards.

```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
...
spec:
  endpoints:
  - relabelings:
    - action: replace
      sourceLabels:
      - __meta_kubernetes_pod_node_name
      targetLabel: nodename
    - action: replace
      sourceLabels:
      - __meta_kubernetes_pod_node_name
      targetLabel: instance
    ...
```

slyt:
[...]

PR #104, opened 2022-09-21
Enable serviceMonitor support prometheus relabelings

kindomLee:


kindomLee:
It may be possible to solve #78

glowkey (dismissed):

kindomLee:
@glowkey Because of signing gpg key by re base, so re-approval required

Thanks

glowkey (approved):

PR #110, opened 2022-10-20
Exit if the hostengine connection goes down

glowkey:
Rather than try to reconnect, simply fail and let controller (k8s, etc) determine whether to restart or not.


dbeer (approved):
Also simple and lgtm

PR #156, opened 2023-04-27
Update go-dcgm bindings

glowkey:


cmd/dcgm-exporter/main.go
```diff
@@ -221,7 +221,9 @@ restart:
 	defer dcgm.FieldsTerm()
 
 	var groups []dcgm.MetricGroup
-	groups, err = dcgm.GetSupportedMetricGroups(0)
+	/* default GPU group is 0 */
```
nikkon-dev, line 224: Technically, this is not true. The group with all GPUs is `DCGM_GROUP_ALL_GPUS (0x7fffffff)`

glowkey: I was referring to the group ID in the output of 'dcgmi group -l'

```
| Groups            |                                                          |
| -> 0              |                                                          |
|    -> Group ID    | 0                                                        |
|    -> Group Name  | DCGM_ALL_SUPPORTED_GPUS                                  |
|    -> Entities    | GPU 0                                                    |
```

PR #219, opened 2023-12-14
Integration tests for the TLS and basic authentication

nvvfedorov:
The PR makes the following changes:

1. The `cmd` package has been refactored to make it testable.
2. Integration tests now test that the dcgm-exporter starts and produces metrics. Also, tests for TLS and basic authentication were added. Integration tests require hardware with a GPU installed for proper execution.
3. Updated Makefile with new commands to run integration tests and linter and check that all dependencies are pushed to Github.
4. Local copies of dependencies are now stored in the vendor folder.


glowkey (commented):
LGTM

glowkey (dismissed):

tests/integration/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+make test-integration -e TEST_ARGS="-test.run [test name here]"
+```
+
+Example:
+```
+make test-integration -e TEST_ARGS="-test.run TestStartAndReadMetrics"
+```
+
+**WARRNING**: It takes about 30 seconds, before the dcgm-exporter instance will read available metrics.
```
nikkon-dev, line 31: Typo: WARNING

nikkon-dev, line 31: Some of the metrics require at least two data points to compute a value, which means at least one polling interval should be passed before we are able to get the results. By default, dcgm-exporter uses 30 seconds polling interval, thus the delay you observe.

nvvfedorov: Done. I have fixed the issue.

nvvfedorov: I updated the readme with this information. @nikkon-dev , thank you.

nikkon-dev (approved):
LGTM

PR #247, opened 2024-02-01
Resolved linter-identified issues for improved code quality

nvvfedorov:
The following issues in the code were resolved:

1. Ignored errors;
2. Code contains unused structs;
3. Some variables don't follow the GOlang naming conventions;
4. Incorrect conditions like: `len(sysInfo.cOpt.MajorRange) <= 0` that causes a functional defects;
5. Loops are unconditionally terminated;
6. Use the deprecated packages and methods.

Also:

1. New unit tests added to increase code coverage by tests;
2. Add a new "--debug" parameter that allows users to see debug messages, which helps to improve troubleshooting.

### DIFF
--- a/pkg/cmd/app.go
+++ b/pkg/cmd/app.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/NVIDIA/dcgm-exporter/pkg/dcgmexporter"
 	"github.com/NVIDIA/go-dcgm/pkg/dcgm"
+	"github.com/davecgh/go-spew/spew"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 )
@@ -63,6 +64,7 @@ const (
 	CLIWebConfigFile            = "web-config-file"
 	CLIXIDCountWindowSize       = "xid-count-window-size"
 	CLIReplaceBlanksInModelName = "replace-blanks-in-model-name"
+	CLIDebugMode                = "debug"
 )
 
 func NewApp(buildVersion ...string) *cli.App {
@@ -190,6 +192,12 @@ func NewApp(buildVersion ...string) *cli.App {
 			Usage:   "Replaces every blank space in the GPU model name with a dash, ensuring a continuous, space-free identifier.",
 			EnvVars: []string{"DCGM_EXPORTER_REPLACE_BLANKS_IN_MODEL_NAME"},
 		},
+		&cli.BoolFlag{
+			Name:    CLIDebugMode,
+			Value:   false,
+			Usage:   "Enable debug output",
+			EnvVars: []string{"DCGM_EXPORTER_DEBUG"},
+		},
 	}
 
 	if runtime.GOOS == "linux" {
@@ -227,6 +235,19 @@ restart:
 	if err != nil {
 		return err
 	}
+
+	if config.Debug {
+		spew.Config.Indent = " "
+		spew.Config.DisablePointerAddresses = true
+		spew.Config.DisableCapacities = true
+		//enable debug logging
+		logrus.SetLevel(logrus.DebugLevel)
+		logrus.Debug("Debug output is enabled")
+	}
+
+	logrus.Debugf("Command line: %s", strings.Join(os.Args, " "))
+
+	logrus.WithField(dcgmexporter.LoggerDumpField, spew.Sdump(config)).Debug("Loaded configuration")
 
 	if config.UseRemoteHE {
 		logrus.Info("Attemping to connect to remote hostengine at ", config.RemoteHEInfo)
@@ -426,5 +447,6 @@ func contextToConfig(c *cli.Context) (*dcgmexporter.Config, error) {
 		WebConfigFile:            c.String(CLIWebConfigFile),
 		XIDCountWindowSize:       c.Int(CLIXIDCountWindowSize),
 		ReplaceBlanksInModelName: c.Bool(CLIReplaceBlanksInModelName),
+		Debug:                    c.Bool(CLIDebugMode),
 	}, nil
 }
--- a/pkg/cmd/app.go
+++ b/pkg/cmd/app.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/NVIDIA/dcgm-exporter/pkg/dcgmexporter"
 	"github.com/NVIDIA/go-dcgm/pkg/dcgm"
-	"github.com/davecgh/go-spew/spew"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 )
@@ -237,9 +236,6 @@ restart:
 	}
 
 	if config.Debug {
-		spew.Config.Indent = " "
-		spew.Config.DisablePointerAddresses = true
-		spew.Config.DisableCapacities = true
 		//enable debug logging
 		logrus.SetLevel(logrus.DebugLevel)
 		logrus.Debug("Debug output is enabled")
@@ -247,7 +243,7 @@ restart:
 
 	logrus.Debugf("Command line: %s", strings.Join(os.Args, " "))
 
-	logrus.WithField(dcgmexporter.LoggerDumpField, spew.Sdump(config)).Debug("Loaded configuration")
+	logrus.WithField(dcgmexporter.LoggerDumpField, fmt.Sprintf("%+v", config)).Debug("Loaded configuration")
 
 	if config.UseRemoteHE {
 		logrus.Info("Attemping to connect to remote hostengine at ", config.RemoteHEInfo)
--- a/pkg/cmd/app.go
+++ b/pkg/cmd/app.go
@@ -243,7 +243,7 @@ restart:
 
 	logrus.Debugf("Command line: %s", strings.Join(os.Args, " "))
 
-	logrus.WithField(dcgmexporter.LoggerDumpField, fmt.Sprintf("%+v", config)).Debug("Loaded configuration")
+	logrus.WithField(dcgmexporter.LoggerDumpKey, fmt.Sprintf("%+v", config)).Debug("Loaded configuration")
 
 	if config.UseRemoteHE {
 		logrus.Info("Attemping to connect to remote hostengine at ", config.RemoteHEInfo)
--- a/pkg/dcgmexporter/const.go
+++ b/pkg/dcgmexporter/const.go
@@ -49,6 +49,11 @@ func mustParseDCGMExporterMetric(s string) DCGMExporterMetric {
 
 // Constants for logging fields
 const (
-	LoggerGroupIDKey = "group_id"
+	LoggerGroupIDKey = "groupID"
 	LoggerDumpKey    = "dump"
+)
+
+const (
+	PARENT_ID_IGNORED      = 0
+	DCGM_ST_NOT_CONFIGURED = "Setting not configured"
 )
--- a/pkg/dcgmexporter/const.go
+++ b/pkg/dcgmexporter/const.go
@@ -46,3 +46,9 @@ func mustParseDCGMExporterMetric(s string) DCGMExporterMetric {
 	}
 	return mv
 }
+
+// Constants for logging fields
+const (
+	LoggerGroupIDKey = "group_id"
+	LoggerDumpKey    = "dump"
+)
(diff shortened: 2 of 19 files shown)


pkg/cmd/app.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	"github.com/NVIDIA/dcgm-exporter/pkg/dcgmexporter"
 	"github.com/NVIDIA/go-dcgm/pkg/dcgm"
+	"github.com/davecgh/go-spew/spew"
```
rohit-arora-dev, line 18: It looks like go-spew has been abandoned and has not been maintained since 2018, do you think it might be better to use some other alternative?

pkg/cmd/app.go
```diff
@@ -228,6 +236,19 @@ restart:
...
+		spew.Config.DisablePointerAddresses = true
+		spew.Config.DisableCapacities = true
+		//enable debug logging
+		logrus.SetLevel(logrus.DebugLevel)
+		logrus.Debug("Debug output is enabled")
+	}
+
+	logrus.Debugf("Command line: %s", strings.Join(os.Args, " "))
+
+	logrus.WithField(dcgmexporter.LoggerDumpField, spew.Sdump(config)).Debug("Loaded configuration")
```
rohit-arora-dev, line 250: One option could be to use `fmt.Sprintf("%+v", config)`.

It changes output from (includes type and len information):
```
dump="(*dcgmexporter.Config)({\n CollectorsFile: (string) (len=26) \"./etc/default-counters.csv\",\n Address: (string) (len=5) \":9400\",\n CollectInterval: (int) 30000,\n Kubernetes: (bool) false,\n KubernetesGPUIdType: (dcgmexporter.KubernetesGPUIDType) (len=3) \"uid\",\n CollectDCP: (bool) true,\n UseOldNamespace: (bool) false,\n UseRemoteHE: (bool) false,\n RemoteHEInfo: (string) (len=14) \"localhost:5555\",\n GPUDevices: (dcgmexporter.DeviceOptions) {\n  Flex: (bool) true,\n  MajorRange: ([]int) <nil>,\n  MinorRange: ([]int) <nil>\n },\n SwitchDevices: (dcgmexporter.DeviceOptions) {\n  Flex: (bool) true,\n  MajorRange: ([]int) <nil>,\n  MinorRange: ([]int) <nil>\n },\n CPUDevices: (dcgmexporter.DeviceOptions) {\n  Flex: (bool) true,\n  MajorRange: ([]int) <nil>,\n  MinorRange: ([]int) <nil>\n },\n NoHostname: (bool) false,\n UseFakeGPUs: (bool) false,\n ConfigMapData: (string) (len=4) \"none\",\n MetricGroups: ([]dcgm.MetricGroup) <nil>,\n WebSystemdSocket: (bool) false,\n WebConfigFile: (string) \"\",\n XIDCountWindowSize: (int) 300000,\n ReplaceBlanksInModelName: (bool) false,\n Debug: (bool) true\n})\n"
```

To (excludes type and len information)::
```
dump="&{CollectorsFile:./etc/default-counters.csv Address::9400 CollectInterval:30000 Kubernetes:false KubernetesGPUIdType:uid CollectDCP:true UseOldNamespace:false UseRemoteHE:false RemoteHEInfo:localhost:5555 GPUDevices:{Flex:true MajorRange:[] MinorRange:[]} SwitchDevices:{Flex:true MajorRange:[] MinorRange:[]} CPUDevices:{Flex:true MajorRange:[] MinorRange:[]} NoHostname:false UseFakeGPUs:false ConfigMapData:none MetricGroups:[] WebSystemdSocket:false WebConfigFile: XIDCountWindowSize:300000 ReplaceBlanksInModelName:false Debug:true}"
```

pkg/dcgmexporter/dcgm.go
```diff
@@ -29,7 +36,14 @@ func NewGroup() (dcgm.GroupHandle, func(), error) {
 		return dcgm.GroupHandle{}, func() {}, err
 	}
 
-	return group, func() { dcgm.DestroyGroup(group) }, nil
+	return group, func() {
+		err := dcgm.DestroyGroup(group)
+		if err != nil {
```
rohit-arora-dev, line 41: Just a nit:
if err := dcgm.DestroyGroup(group); err != nil`

Thanks for handling the return error.

pkg/cmd/app.go
nvvfedorov: I agree, at the same time, Spew is a very popular package; for example, K8S uses it.

pkg/dcgmexporter/dcgm.go
nvvfedorov: This is the point of the PR :) Linter detects such issues :)

pkg/cmd/app.go
nvvfedorov: I removed spew usage.

pkg/cmd/app.go
nvvfedorov: done.

rohit-arora-dev (commented):
@nvvfedorov Thanks for this change and making significant improvements. I tend to leave detailed comments and lots of `nits`. If you feel some of them are not applicable we can definitly discuss as I am still coming up to speed on dcgm-exporter and go-dcgm and would be an opportunity for me to gain some context.

As a long term I feel we should explore moving logging to its own package and support various logging levels along with some context information for each request.

pkg/cmd/app.go
rohit-arora-dev: Thank you!

pkg/dcgmexporter/system_info.go
```diff
@@ -461,25 +468,33 @@ func CreateCoreGroupsFromSystemInfo(sysInfo SystemInfo) ([]dcgm.GroupHandle, []f
...
+			err = dcgm.AddEntityToGroup(groupID, dcgm.FE_CPU_CORE, core)
 
 			if err != nil {
 				return groups, cleanups, err
 			}
 
-			cleanups = append(cleanups, func() { dcgm.DestroyGroup(groupId) })
+			cleanups = append(cleanups, func() {
+				err := dcgm.DestroyGroup(groupID)
+				if err != nil && !strings.Contains(err.Error(), DCGM_ST_NOT_CONFIGURED) {
```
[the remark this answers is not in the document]
nvvfedorov: You are right. Today, go-dcgm returns an error. We may change go-dcgm, to return a custom return to make it possible to use the "errors.As" to extract error code. We will do this later, outside of the current PR.

pkg/dcgmexporter/system_info.go
```diff
@@ -618,87 +647,92 @@ func AddAllLinks(sysInfo SystemInfo) []MonitoringInfo {
...
 
-func IsSwitchWatched(switchId uint, sysInfo SystemInfo) bool {
+func IsSwitchWatched(switchID uint, sysInfo SystemInfo) bool {
 	if sysInfo.sOpt.Flex {
 		return true
 	}
 
-	if len(sysInfo.sOpt.MajorRange) <= 0 {
+	// When MajorRange contains -1 value, we do monitorig of all switches
+	if len(sysInfo.sOpt.MajorRange) > 0 && sysInfo.sOpt.MajorRange[0] == -1 {
```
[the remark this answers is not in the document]
nvvfedorov: The "-1" value means that we monitor all devices. See here: https://github.com/NVIDIA/dcgm-exporter/blob/main/pkg/cmd/app.go#L349.

pkg/dcgmexporter/system_info.go
```diff
@@ -618,87 +647,92 @@ func AddAllLinks(sysInfo SystemInfo) []MonitoringInfo {
...
-	for _, sw := range sysInfo.Switches {
-		if !IsSwitchWatched(sw.EntityId, sysInfo) {
-			return false
-		}
+	// Find a switch
+	switchIdx := slices.IndexFunc(sysInfo.Switches, func(si SwitchInfo) bool {
+		return IsSwitchWatched(si.EntityId, sysInfo) && si.EntityId == switchID
+	})
+
+	if switchIdx > -1 {
```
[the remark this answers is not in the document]
nvvfedorov: The logic of the function is the following: 

When sysInfo.sOpt.Flex == true - we watch all links on all switches;

Otherwise: 

1. We check if the switch associated with the monitored link is watched. The list of switches contains unique values.
2. We check if the link id is a list of monitored links. 

When both conditions are TRUE, we return TRUE.

pkg/dcgmexporter/system_info.go
```diff
@@ -618,87 +647,92 @@ func AddAllLinks(sysInfo SystemInfo) []MonitoringInfo {
...
-			if uint(link) == linkId {
-				return true
-			}
+		// The Link exists
+		if slices.ContainsFunc(sw.NvLinks, func(nls dcgm.NvLinkStatus) bool {
+			return nls.Index == linkIndex
+		}) {
+			// and the link index in the Minor range
+			return slices.ContainsFunc(sysInfo.sOpt.MinorRange, func(i int) bool {
+				return uint(i) == linkIndex
```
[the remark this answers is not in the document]
nvvfedorov: Thanks

pkg/dcgmexporter/system_info.go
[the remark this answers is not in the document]
rohit-arora-dev: Thanks for clarifying, it looks like there was a bug in the existing implementation, missing the association between the switch and the link.

nit: Changing `IsSwitchWatched(si.EntityId, sysInfo) && si.EntityId == switchID` to `IsSwitchWatched(si.EntityId == switchID && si.EntityId, sysInfo)` can avoid some of the function call.

pkg/dcgmexporter/server.go
```diff
@@ -50,13 +50,17 @@ func NewMetricsServer(c *Config, metrics chan string, registry *Registry) (*Metr
...
-		w.Write([]byte(`<html>
+		_, err := w.Write([]byte(`<html>
 			<head><title>GPU Exporter</title></head>
 			<body>
 			<h1>GPU Exporter</h1>
 			<p><a href="./metrics">Metrics</a></p>
 			</body>
 			</html>`))
+		if err != nil {
+			http.Error(w, "Failed to write response", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
nvvfedorov: I think, adding logging will be better than writing an error message to the response. OWASP suggests avoiding returning internal details in the web app response. See https://cheatsheetseries.owasp.org/cheatsheets/Error_Handling_Cheat_Sheet.html.

pkg/dcgmexporter/system_info.go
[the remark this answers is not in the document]
nvvfedorov: Good point!

pkg/dcgmexporter/system_info.go
```diff
@@ -19,13 +19,16 @@ package dcgmexporter
...
+	"slices"
+	"strings"
 
 	"github.com/NVIDIA/go-dcgm/pkg/dcgm"
 	"github.com/bits-and-blooms/bitset"
 	"github.com/sirupsen/logrus"
 )
 
 const PARENT_ID_IGNORED = 0
+const DCGM_ST_NOT_CONFIGURED = "Setting not configured"
```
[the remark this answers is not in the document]
rohit-arora-dev: Lets fix this as part of another PR.

pkg/dcgmexporter/server.go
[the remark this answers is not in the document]
rohit-arora-dev: Thanks for adding logs, that provides a way to the end user to identify the underlying issue.

pkg/dcgmexporter/const.go
```diff
@@ -46,3 +46,9 @@ func mustParseDCGMExporterMetric(s string) DCGMExporterMetric {
 	}
 	return mv
 }
+
+// Constants for logging fields
+const (
+	LoggerGroupIDKey = "group_id"
```
rohit-arora-dev, line 52: I still strongly believe we should follow Go convention here and use mixedCaps since that's the Go convention.

nvvfedorov: I made suggested changes.

rohit-arora-dev (approved):
@nvvfedorov This is definitely a really good improvement, thanks for taking this initiative and addressing the review comments.

/LGTM

Issue #250, opened 2024-02-08
`dcgm-exporter` on `main` does not report metrics for all GPUs 


mahendrapaipuri:
Hello,

The latest version (`main` branch) of `dcgm-exporter` is not reporting metrics of all GPUs. I think the exporter regressed since adding the support for Hopper CPU. With a older version (`3.3.0-3.2.0`), it was working as expected.

`main` branch:

- Compiled `dcgm-exporter` using `make binary` command with Go 1.21.3
- Started exporter using `dcgm-exporter -f counters.csv -d g:0,1,2,3`
- Logs:

```
INFO[0000] Starting dcgm-exporter                       
INFO[0000] DCGM successfully initialized!               
INFO[0000] Collecting DCP Metrics                       
INFO[0000] No configmap data specified, falling back to metric file service_files/default-counters.csv 
INFO[0000] Initializing system entities of type: GPU    
INFO[0000] Not collecting switch metrics: No fields to watch for device type: 3 
INFO[0000] Not collecting link metrics: No fields to watch for device type: 6 
INFO[0000] Not collecting cpu metrics: No fields to watch for device type: 7 
INFO[0000] Not collecting cpu core metrics: No fields to watch for device type: 8 
INFO[0000] Pipeline starting                            
INFO[0000] Starting webserver                           
level=info ts=2024-02-08T09:13:41.890Z caller=tls_config.go:313 msg="Listening on" address=[::]:9400
level=info ts=2024-02-08T09:13:41.890Z caller=tls_config.go:316 msg="TLS is disabled." http2=false address=[::]:9400
```
- Output:

```
# HELP DCGM_FI_DEV_SM_CLOCK SM clock frequency (in MHz).
# TYPE DCGM_FI_DEV_SM_CLOCK gauge
DCGM_FI_DEV_SM_CLOCK{gpu="3",UUID="GPU-bd94f0f5-8c2f-bc95-74ad-5f99fdc33d01",device="nvidia3",modelName="Tesla V100-SXM2-32GB",Hostname="r6i3n0"} 135
# HELP DCGM_FI_DEV_MEM_CLOCK Memory clock frequency (in MHz).
# TYPE DCGM_FI_DEV_MEM_CLOCK gauge
[...]

PR #251, opened 2024-02-08
The issue "250 dcgm-exporter on main does not report metrics for all GPUs"

nvvfedorov:
Bug fix for the issue #250 dcgm-exporter on main does not report metrics for all GPUs.

### DIFF
--- a/pkg/dcgmexporter/gpu_collector.go
+++ b/pkg/dcgmexporter/gpu_collector.go
@@ -90,9 +90,8 @@ func (c *DCGMCollector) Cleanup() {
 
 func (c *DCGMCollector) GetMetrics() (map[Counter][]Metric, error) {
 	monitoringInfo := GetMonitoredEntities(c.SysInfo)
-	count := len(monitoringInfo)
 
-	metrics := make(map[Counter][]Metric, count)
+	metrics := make(map[Counter][]Metric)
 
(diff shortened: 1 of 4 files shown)


glowkey (approved):

pkg/dcgmexporter/gpu_collector.go
```diff
@@ -90,9 +90,8 @@ func (c *DCGMCollector) Cleanup() {
 
 func (c *DCGMCollector) GetMetrics() (map[Counter][]Metric, error) {
 	monitoringInfo := GetMonitoredEntities(c.SysInfo)
-	count := len(monitoringInfo)
 
-	metrics := make(map[Counter][]Metric, count)
+	metrics := make(map[Counter][]Metric)
```
rohit-arora-dev, line 94: Just an optional suggestion - may be declare `type GPUMetrics map[Counter][]Metric` in `pkg/dcgmexporter/types.go`, declaring the type would improve readability and optionally provide an opportunity to define methods in future if needed.

rohit-arora-dev (approved):
/LGTM

Thanks for fixing this and adding UTs.

pkg/dcgmexporter/gpu_collector.go
nvvfedorov: I will make this change in another PR.

PR #253, opened 2024-02-14
Make error message styling consistent

rohit-arora-dev:
The purpose of this PR is to achieve consistent error message styling. Styling follows the following conventions:

1. Error strings should not be capitalized (unless beginning with an exported name, a proper noun or an acronym) and should not end with punctuation. 
```
err := fmt.Errorf("something bad happened")
```

2. Error message is separated from the main message string using `;` seperator.
```
err := fmt.Errorf("something bad happened; err: %w", err)
```

3. Use of `%w` in place of `%v` to make it available to `errors.Is` and `errors.As` when supporting custom errors in future.

**NOTE:** There were 2-3 instances where it was not possible to applying the correct styling without changing existing logic a bit. So please pay special attention to those changes.

### DIFF
--- a/internal/pkg/nvmlprovider/provider.go
+++ b/internal/pkg/nvmlprovider/provider.go
@@ -43,7 +43,7 @@ func GetMIGDeviceInfoByID(uuid string) (*MIGDeviceInfo, error) {
 		ret := nvml.Init()
 		if ret != nvml.SUCCESS {
 			err = errors.New(nvml.ErrorString(ret))
-			logrus.Error("Can not init NVML library")
+			logrus.Error("Can not init NVML library.")
 		}
 	})
 	if err != nil {
@@ -88,22 +88,22 @@ func GetMIGDeviceInfoByID(uuid string) (*MIGDeviceInfo, error) {
 
 	tokens := strings.SplitN(uuid, "-", 2)
 	if len(tokens) != 2 || tokens[0] != "MIG" {
-		return nil, fmt.Errorf("Unable to parse UUID as MIG device")
+		return nil, fmt.Errorf("unable to parse UUID '%s' as MIG device", uuid)
 	}
 
 	tokens = strings.SplitN(tokens[1], "/", 3)
 	if len(tokens) != 3 || !strings.HasPrefix(tokens[0], "GPU-") {
-		return nil, fmt.Errorf("Unable to parse UUID as MIG device")
+		return nil, fmt.Errorf("unable to parse UUID '%s' as MIG device", uuid)
 	}
 
 	gi, err := strconv.Atoi(tokens[1])
 	if err != nil {
-		return nil, fmt.Errorf("Unable to parse UUID as MIG device")
+		return nil, fmt.Errorf("unable to parse UUID '%s' as MIG device", uuid)
 	}
 
 	ci, err := strconv.Atoi(tokens[2])
 	if err != nil {
-		return nil, fmt.Errorf("Unable to parse UUID as MIG device")
+		return nil, fmt.Errorf("unable to parse UUID '%s' as MIG device", uuid)
 	}
 
 	return &MIGDeviceInfo{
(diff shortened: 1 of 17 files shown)


internal/pkg/nvmlprovider/provider.go
```diff
@@ -88,22 +88,22 @@ func GetMIGDeviceInfoByID(uuid string) (*MIGDeviceInfo, error) {
 
 	tokens := strings.SplitN(uuid, "-", 2)
 	if len(tokens) != 2 || tokens[0] != "MIG" {
-		return nil, fmt.Errorf("Unable to parse UUID as MIG device")
+		return nil, fmt.Errorf("unable to parse UUID '%s' as MIG device", uuid)
```
rohit-arora-dev, line 91: This is an optional change to enhance message, if required this can be reverted.

pkg/dcgmexporter/parser.go
```diff
@@ -52,15 +52,15 @@ func ExtractCounters(c *Config) ([]Counter, []Counter, error) {
...
 		}
 	} else {
-		err = fmt.Errorf("no configmap data specified")
+		logrus.Infof("No configmap data specified")
 	}
 
-	if err != nil {
-		logrus.Infof("%v, falling back to metric file %s", err, c.CollectorsFile)
+	if err != nil || c.ConfigMapData == undefinedConfigMapData {
+		logrus.Infof("Falling back to metric file '%s'", c.CollectorsFile)
```
rohit-arora-dev, line 59: When reviewing @glowkey and @nvvfedorov please pay special attention to this change.

pkg/dcgmexporter/system_info.go
```diff
@@ -97,21 +97,22 @@ func SetGPUInstanceProfileName(sysInfo *SystemInfo, entityId uint, profileName s
...
+	errStr := fmt.Sprintf("cannot find match for entities:")
+
 	for _, v := range values {
-		found := SetGPUInstanceProfileName(sysInfo, v.EntityId, dcgm.Fv2_String(v))
-		if !found {
-			err = fmt.Errorf("%s group %d, id %d", err, v.EntityGroupId, v.EntityId)
-			notFound = true
+		if !SetGPUInstanceProfileName(sysInfo, v.EntityId, dcgm.Fv2_String(v)) {
+			errStr = fmt.Sprintf("%s group %d, id %d", errStr, v.EntityGroupId, v.EntityId)
+			errFound = true
```
rohit-arora-dev, line 107: When reviewing @glowkey and @nvvfedorov please pay special attention to this change, if this improvement is ok as part of this change.

pkg/dcgmexporter/parser.go
nvvfedorov: We need to add a unit test for this part of the code.

pkg/dcgmexporter/system_info.go
nvvfedorov: We need to add a unit test for this part of the code.

nvvfedorov (changes requested):
Please add unit tests to cover changed lines of code.

pkg/dcgmexporter/parser.go
rohit-arora-dev: The existing test case `TestExtractCounters` already covers this part of the code where it reads from the `prefix-` file created by the test case and K8s config map is not supplies. 

Unless you have a specific test case in mind.

pkg/dcgmexporter/system_info.go
rohit-arora-dev: Adding, thanks for identifying the gap.

rohit-arora-dev: @nvvfedorov Added UT for `SetMigProfileNames`.

nvvfedorov (approved):
LGTM

PR #258, opened 2024-02-21
Update Go Dependencies

rohit-arora-dev:
The purpose of this change is to update Golang dependencies. As part of this change `CreateListener` dependent on K8s utils in `pkg/dcgmexporter/kubernetes_test.go` file has been replaced by net package, thus reducing a number dependencies.


nvvfedorov:
> The purpose of this change is to update Golang dependencies. As part of this change `CreateListener` dependent on K8s utils in `pkg/dcgmexporter/kubernetes_test.go` file has been replaced by net package, thus reducing a number dependencies.

Good change!

nvvfedorov (approved):
LGTM

PR #276, opened 2024-03-01
Added ability to enable DCGM library log output

nvvfedorov:
Users are now able to use the "--enable-dcgm-log" and "--dcgm-log-level=[NONE,FATAL,ERROR,WARN,INFO,DEBUG,VERB]" flags to enable DCGM library log output into the dcgm-exporter logs. 

**Important:** The default value for the "--dcgm-log-level" is "NONE". To see DCGM logs, you must set at least the `INFO` log level.

Here is an example of the DCGM log output:

```
time="2024-03-01T14:24:35-06:00" level=info msg="version:3.3.3;arch:x86_64;buildtype:Release;buildid:11;builddate:2024-01-18;commit:c3aed64480553cd5ba1a32d165c7967936446631;branch:rel_dcgm_3_3;buildplatform:Linux 4.15.0-180-generic #189-Ubuntu SMP Wed May 18 14:13:57 UTC 2022 x86_64;;crc:39ff792f5514bdc5af56ce313a8fa90e [/workspaces/dcgm-rel_dcgm_3_3-postmerge/dcgmlib/src/DcgmApi.cpp:5012] [{anonymous}::StartEmbeddedV2]" dcgm_level=INFO
time="2024-03-01T14:24:35-06:00" level=info msg="Signal 12 is already handled. Nothing to do. [/workspaces/dcgm-rel_dcgm_3_3-postmerge/common/DcgmThread/DcgmThread.cpp:394] [DcgmThread::InstallSignalHandler]" dcgm_level=INFO
time="2024-03-01T14:24:35-06:00" level=info msg="Parsed driver string is 5452902, IsR450OrNewer: 1, IsR520OrNewer: 1 [/workspaces/dcgm-rel_dcgm_3_3-postmerge/dcgmlib/src/DcgmCacheManager.cpp:2592] [DcgmCacheManager::ReadAndCacheDriverVersions]" dcgm_level=INFO
time="2024-03-01T14:24:35-06:00" level=info msg="Detected 0 NVLinks for GPU 0 [/workspaces/dcgm-rel_dcgm_3_3-postmerge/dcgmlib/src/DcgmCacheManager.cpp:1263] [DcgmCacheManager::InitializeNvLinkCount]" dcgm_level=INFO
time="2024-03-01T14:24:35-06:00" level=info msg="Got 0 excluded GPUs [/workspaces/dcgm-rel_dcgm_3_3-postmerge/dcgmlib/src/DcgmCacheManager.cpp:1087] [DcgmCacheManager::ReadAndCacheGpuExclusionList]" dcgm_level=INFO
```

### DIFF
--- a/internal/pkg/logging/logger_adapter.go
+++ b/internal/pkg/logging/logger_adapter.go
@@ -1,0 +1,75 @@
+/*
+ * Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package logging
+
+import (
+	"fmt"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/sirupsen/logrus"
+)
+
+// LogrusAdapter is an adapter that allows logrus Logger to be used as a go-kit/log Logger.
+type LogrusAdapter struct {
+	Logger *logrus.Logger
+}
+
+// NewLogrusAdapter creates a new LogrusAdapter with the provided logrus.Logger.
+func NewLogrusAdapter(logger *logrus.Logger) log.Logger {
+	return &LogrusAdapter{
+		Logger: logger,
+	}
+}
+
+// Log implements the go-kit/log Logger interface.
+func (la *LogrusAdapter) Log(keyvals ...interface{}) error {
+	// keyvals is a slice of interfaces, that represents a key-value pairs.
+	if len(keyvals)%2 != 0 {
+		keyvals = append(keyvals, "MISSING")
+	}
+
+	fields := logrus.Fields{}
+	for i := 0; i < len(keyvals); i += 2 {
+		key, ok := keyvals[i].(string)
+		if !ok {
+			// If the key is not la string, use la default key
+			key = "missing_key"
+		}
+		fields[key] = keyvals[i+1]
+	}
+
+	// The go-kit/log uses msg field to keep log message, we don't want to use message as field in the logrus.
+	msg, exists := fields["msg"]
+	if exists {
+		delete(fields, "msg")
+	}
+
+	// The go-kit/log uses level fields to keep log level. We need to convert this field into logrus value.
+	lvl, exists := fields["level"]
+	if !exists {
+		fields["level"] = level.InfoValue()
+	}
+	delete(fields, "level")
+	parsedLvl, err := logrus.ParseLevel(fmt.Sprint(lvl))
+	if err != nil {
+		parsedLvl = logrus.InfoLevel
+	}
+
+	la.Logger.WithFields(fields).Log(parsedLvl, msg)
+
+	return nil
+}
--- a/pkg/cmd/app.go
+++ b/pkg/cmd/app.go
@@ -2,11 +2,13 @@ package cmd
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"os"
 	"os/signal"
 	"runtime"
 	"runtime/debug"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -14,11 +16,11 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/NVIDIA/dcgm-exporter/pkg/dcgmexporter"
+	"github.com/NVIDIA/dcgm-exporter/pkg/stdout"
 	"github.com/NVIDIA/go-dcgm/pkg/dcgm"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
-
-	"github.com/NVIDIA/dcgm-exporter/pkg/dcgmexporter"
 )
 
 const (
@@ -67,6 +69,8 @@ const (
 	CLIReplaceBlanksInModelName   = "replace-blanks-in-model-name"
 	CLIDebugMode                  = "debug"
 	CLIClockEventsCountWindowSize = "clock-events-count-window-size"
+	CLIEnableDCGMLog              = "enable-dcgm-log"
+	CLIDCGMLogLevel               = "dcgm-log-level"
 )
 
 func NewApp(buildVersion ...string) *cli.App {
@@ -191,7 +195,7 @@ func NewApp(buildVersion ...string) *cli.App {
 			Name:    CLIReplaceBlanksInModelName,
 			Aliases: []string{"rbmn"},
 			Value:   false,
-			Usage:   "Replaces every blank space in the GPU model name with a dash, ensuring a continuous, space-free identifier.",
+			Usage:   "Replace every blank space in the GPU model name with a dash, ensuring a continuous, space-free identifier.",
 			EnvVars: []string{"DCGM_EXPORTER_REPLACE_BLANKS_IN_MODEL_NAME"},
 		},
 		&cli.BoolFlag{
@@ -205,6 +209,18 @@ func NewApp(buildVersion ...string) *cli.App {
 			Value:   int((5 * time.Minute).Milliseconds()),
 			Usage:   "Set time window size in milliseconds (ms) for counting clock events in DCGM Exporter.",
 			EnvVars: []string{"DCGM_EXPORTER_CLOCK_EVENTS_COUNT_WINDOW_SIZE"},
+		},
+		&cli.BoolFlag{
+			Name:    CLIEnableDCGMLog,
+			Value:   false,
+			Usage:   "Enable writing DCGM logs to standard output (stdout).",
+			EnvVars: []string{"DCGM_EXPORTER_ENABLE_DCGM_LOG"},
+		},
+		&cli.StringFlag{
+			Name:    CLIDCGMLogLevel,
+			Value:   dcgmexporter.DCGMDbgLvlNone,
+			Usage:   "Specify the DCGM log verbosity level. This parameter is effective only when the '--enable-dcgm-log' option is set to 'true'. Possible values: NONE, FATAL, ERROR, WARN, INFO, DEBUG and VERB",
+			EnvVars: []string{"DCGM_EXPORTER_DCGM_LOG_LEVEL"},
 		},
 	}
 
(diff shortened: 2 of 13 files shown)


pkg/stdout/logprocessor.go
```diff
@@ -0,0 +1,62 @@
...
+// parseLogEntry takes a log entry string and returns a structured LogEntry object.
+func parseLogEntry(entry string) LogEntry {
+	// Split the entry by spaces, taking care to not split the function call and its arguments.
+	fields := strings.Fields(entry)
+
+	if len(fields) > 2 {
+		// Parse the timestamp.
+		timestamp, err := time.Parse("2006-01-02 15:04:05.000", fields[0]+" "+fields[1])
+		if err != nil {
+			return LogEntry{
```
rohit-arora-dev, line 40: Please add a log statement explaining the failure.

pkg/stdout/capture.go
```diff
@@ -0,0 +1,105 @@
...
+	"bufio"
+	"context"
+	"log"
+	"os"
+	"syscall"
+
+	"github.com/sirupsen/logrus"
+)
+
+func Capture(ctx context.Context, inner func() error) error {
```
rohit-arora-dev, line 29: Please add explanation of the func `Capture` in Comment above.

pkg/stdout/capture.go
```diff
@@ -0,0 +1,105 @@
...
+
+	defer func() {
+		cleanup()
+	}()
+
+	// Call function here
+	return inner()
+}
+
+func mustHijackStdOut() (r *os.File, o *os.File, cleanup func()) {
```
rohit-arora-dev, line 61: Please add explanation of the func `mustHijackStdOut` in Comment above.

Naming preference: `captureStdOut` instead of `mustHijackStdOut`

pkg/stdout/capture.go
```diff
@@ -0,0 +1,105 @@
...
+	"context"
+	"log"
+	"os"
+	"syscall"
+
+	"github.com/sirupsen/logrus"
+)
+
+func Capture(ctx context.Context, inner func() error) error {
+	r, _, cleanup := mustHijackStdOut()
```
rohit-arora-dev, line 30: No need to capture `stdErr`?

internal/pkg/logging/logger_adapter.go
```diff
@@ -0,0 +1,55 @@
...
+
+// NewLogrusAdapter creates a new LogrusAdapter with the provided logrus.Logger.
+func NewLogrusAdapter(logger *logrus.Logger) log.Logger {
+	return &LogrusAdapter{
+		Logger: logger,
+	}
+}
+
+// Log implements the go-kit/log Logger interface.
+func (a *LogrusAdapter) Log(keyvals ...interface{}) error {
```
rohit-arora-dev, line 37: nit:
`la *LogrusAdapter` or `l *LogrusAdapter`

pkg/dcgmexporter/const.go
```diff
@@ -31,3 +31,22 @@ const (
...
+
+const (
+	DCGMDbgLvlNone  = "NONE"
+	DCGMDbgLvlFatal = "FATAL"
+	DCGMDbgLvlError = "ERROR"
+	DCGMDbgLvlWarn  = "WARN"
+	DCGMDbgLvlInfo  = "INFO"
+	DCGMDbgLvlDebug = "DEBUG"
+	DCGMDbgLvlVerb  = "VERB"
+)
```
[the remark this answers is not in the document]
nvvfedorov: The purpose of the such names, to tell that mentioned debug level constants belongs to the DCGM Library and not to the dcgm-exporter.  Available names defined here: https://github.com/NVIDIA/DCGM/blob/18b87c715750d0d44b185dfeb9a7d8e2597443a4/common/DcgmLogging.h#L35C1-L36C1

pkg/stdout/logprocessor.go
```diff
@@ -0,0 +1,62 @@
...
+// LogEntry represents the structured form of the parsed log entry.
+type LogEntry struct {
+	Timestamp   time.Time
+	Level       string
+	Message     string
+	IsRawString bool
+}
+
+// parseLogEntry takes a log entry string and returns a structured LogEntry object.
+func parseLogEntry(entry string) LogEntry {
```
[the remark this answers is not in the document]
nvvfedorov: @rohit-arora-dev  What are the UTs?

pkg/stdout/capture.go
nvvfedorov: No, today we don't need.

pkg/cmd/app.go
```diff
@@ -236,180 +252,188 @@ func newOSWatcher(sigs ...os.Signal) chan os.Signal {
...
-		// enable debug logging
-		logrus.SetLevel(logrus.DebugLevel)
-		logrus.Debug("Debug output is enabled")
-	}
-
-	logrus.Debugf("Command line: %s", strings.Join(os.Args, " "))
-
-	logrus.WithField(dcgmexporter.LoggerDumpKey, fmt.Sprintf("%+v", config)).Debug("Loaded configuration")
+	ctx, cancel := context.WithCancel(context.Background())
+	return stdout.Capture(ctx, func() error {
```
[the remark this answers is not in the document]
nvvfedorov: @rohit-arora-dev , What is the HoF?

pkg/stdout/logprocessor.go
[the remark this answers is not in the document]
rohit-arora-dev: Unit test

pkg/cmd/app.go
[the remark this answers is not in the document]
rohit-arora-dev: Higher Order Functions

pkg/dcgmexporter/const.go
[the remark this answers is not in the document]
rohit-arora-dev: If exclusive to DCGM then make sense to keep the prefix.

pkg/stdout/capture.go
rohit-arora-dev: Would it not help identify any kind of failure in exporter if there is an issue in dcgm?

pkg/cmd/app.go
```diff
@@ -514,5 +542,14 @@ func contextToConfig(c *cli.Context) (*dcgmexporter.Config, error) {
...
+		DCGMLogLevel:               dcgmLogLevel,
 	}, nil
 }
+
+// LogEntry represents the structured form of the parsed log entry.
+type LogEntry struct {
+	Timestamp time.Time
+	Level     string
+	Message   string
+}
```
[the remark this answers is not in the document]
nvvfedorov: @rohit-arora-dev , This is a good catch. The structure is a left over from the refactoring and must be removed.

pkg/stdout/logprocessor.go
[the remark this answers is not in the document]
nvvfedorov: The func is covered by the test: https://github.com/NVIDIA/dcgm-exporter/blob/enable-dcgm-logging-options/pkg/stdout/capture_test.go#L31.

pkg/stdout/capture.go
nvvfedorov: No. Today it is not on the table for the change.

pkg/cmd/app.go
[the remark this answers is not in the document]
nvvfedorov: The HoS doesn't have any impact on testability. However, implementation without HoS will significantly increase the complexity of the capturer.

pkg/cmd/app.go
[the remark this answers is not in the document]
nvvfedorov: Done. I remove the structure.

pkg/stdout/logprocessor.go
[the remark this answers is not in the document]
nvvfedorov: Unit tests were added.

pkg/cmd/app.go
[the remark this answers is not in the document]
nvvfedorov: Unfortunately, the proposed approach with Start/Stop doesn't work well for the use case when we capture GO and C stdout because of the global nature of the stdout.  

The GO runtime and C runtime have different implementations of the `STDOUT` and `IO` implementations.  The `os.Stdout` is not the same file that C uses when interacting with `syscall.Stdout`. When we create a pipe to read and forward `syscall.Stdout`, we get the `syscall.Stdout` file descriptor and replace it with a file descriptor of the pipe writer. After this, ALL writes to `/dev/stdout` inside the process go to the pipe and processing. If we run the code as part of the unit test, the unit test framework output will also go the pipe. That is why restoring the process's original state is critical. The implemented functional approach ensure that changes in `syscall.Stdout` doesn't escape of the function implementation. Also, the usage of `syscall` is unsafe code, meaning it is runs outside of the GO runtime and GO garbage collector cleans the stack after function execution. 

All these things make it difficult to implement the suggested Start/Stop approach, as it has side effects.

pkg/stdout/capture.go
nvvfedorov: Done

internal/pkg/logging/logger_adapter.go
nvvfedorov: done

internal/pkg/logging/logger_adapter.go
```diff
@@ -0,0 +1,72 @@
...
+
+	msg, exists := fields["msg"]
+	if exists {
+		delete(fields, "msg")
+	}
+
+	lvl, exists := fields["level"]
+	if !exists {
+		fields["level"] = level.InfoValue()
+	}
```
rohit-arora-dev, line 62: nit:
Would appreciate if you can add a comment explaning this new change.
And add a new linf between 63 and 64.

pkg/cmd/app.go
```diff
@@ -346,22 +310,49 @@ restart:
...
+
+	var wg sync.WaitGroup
+	stop := make(chan interface{})
+
+	wg.Add(1)
+	go pipeline.Run(ch, stop, &wg)
+
+	wg.Add(1)
+
+	server, cleanup, err := dcgmexporter.NewMetricsServer(config, ch, cRegistry)
```
rohit-arora-dev, line 331: I see some change in the ordering, is that on purpose? Should this be before line 321?

pkg/cmd/app.go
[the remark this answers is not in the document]
rohit-arora-dev: Thanks for giving it a try.

rohit-arora-dev (commented):
Just couple of minor suggestions based on your recent changes otherwise looking good. Thanks for addressing the review comments.

pkg/stdout/capture.go
nvvfedorov: done

pkg/cmd/app.go
nvvfedorov: @rohit-arora-dev, Yes, the change in the ordering was made to move variable assignments closer to the first usage.

internal/pkg/logging/logger_adapter.go
nvvfedorov: @rohit-arora-dev , done.

pkg/stdout/logprocessor.go
nvvfedorov: There is no need to return the error here. Instead of logging an error, if we can not parse the log, we use the log string as is.

rohit-arora-dev (approved):
Thanks @nvvfedorov for this change and addressing the review comments.

/LGTM

PR #285, opened 2024-03-14
Issue-272: Added tests for K8S use case, when MIG enabled in a mixed mode

nvvfedorov:
The new test case was added to capture the use case when the GPU was configured to operate in MIG mode.


glowkey (approved):
LGTM

PR #288, opened 2024-03-18
Fix power calculation

Apsu:
Using the calculation of `sum` for the power gauge results in summing the gauge over time, rather than reflecting the average amount for the time window. Swap it for `lastNotNull`.

### DIFF
--- a/grafana/dcgm-exporter-dashboard.json
+++ b/grafana/dcgm-exporter-dashboard.json
@@ -304,7 +304,7 @@
       "options": {
         "fieldOptions": {
           "calcs": [
-            "sum"
+            "lastNotNull"
           ],
           "defaults": {
             "color": {


nvvfedorov:
@Apsu, Thank you for the PR. Can you provide examples before/after the change and justify the change?

frittentheke:
@Apsu the panel already used `instant` (https://grafana.com/docs/grafana/latest/datasources/prometheus/query-editor/#type) causing only the most recent value to be used.

I believe if you saw issue that is / was due to e.g. multiple devices (MIG?) being summed together?
See my issue https://github.com/NVIDIA/dcgm-exporter/issues/353 and PR https://github.com/NVIDIA/dcgm-exporter/pull/355

nccurry:
Closing this PR as stale. 
@Apsu We can re-open this if you feel this is still needed and the suggested change is still correct in the latest version of DCGM-Exporter. 

Thanks!

PR #295, opened 2024-03-20
E2E tests can be run with default-counters.csv

nvvfedorov:
The E2E tests with default configuration use the helm release, which deploys the dcgm-exporter with the `dcp-metrics-included.csv` configuration. This configuration includes profiling metrics available on the datacenter-grade GPU.  
For example, if you use `NVIDIA T400`, the e2e test fails because the dcgm-exporter can not be initiated because of unsupported hardware. 

To provide the ability to run e2e tests on low-grade GPUs compatible with DCGM, I added the ability to specify custom arguments in the e2e scripts. As a result:
1. The default `make e2e` command runs e2e tests with enabled profiling metrics.
2. If the developer wants to run the e2e tests without profiling metrics, they can do the following:

```
cd tests/e2e
make e2e-test-no-profiling
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,6 +5,16 @@
     "version": "0.2.0",
     "configurations": [
         {
+            "name": "Go: Launch e2e",
+            "type": "go",
+            "request": "launch",
+            "mode": "test",
+            "program": "${workspaceFolder}/tests/e2e",
+            "args": ["-kubeconfig","~/.kube/config", "-chart","./../../deployment/","-image-repository","nvidia/dcgm-exporter","-arguments","{-f=/etc/dcgm-exporter/dcp-metrics-included.csv,--enable-dcgm-log=true,--dcgm-log-level=ERROR}"],
+            "env": {},
+            "buildFlags": "-tags=e2e"
+        },
+        {
             "name": "Run Debug",
             "type": "go",
             "request": "launch",
--- a/tests/e2e/Makefile
+++ b/tests/e2e/Makefile
@@ -25,9 +25,25 @@ e2e-test:
 		echo "[ERR] KUBECONFIG is missing, must be set"; \
 		exit 1; \
 	fi
-	$(GO_CMD) test --tags=e2e -v . -args \
+	$(GO_CMD) test -failfast --tags=e2e -v . -args \
 		-kubeconfig=$(KUBECONFIG) \
 		-chart="$(CHART)" \
 		-namespace=$(NAMESPACE) \
 		-image-repository=$(IMAGE_REPOSITORY) \
 		-image-tag=$(IMAGE_TAG)
+
+
+.PHONY: e2e-test-no-profiling
+e2e-test-no-profiling:
+	@if [ -z ${KUBECONFIG} ]; then \
+		echo "[ERR] KUBECONFIG is missing, must be set"; \
+		exit 1; \
+	fi
+	$(GO_CMD) test -failfast --tags=e2e -v . -args \
+		-kubeconfig=$(KUBECONFIG) \
+		-chart="$(CHART)" \
+		-namespace=$(NAMESPACE) \
+		-image-repository=$(IMAGE_REPOSITORY) \
+		-image-tag=$(IMAGE_TAG) \
+		-arguments="{-f=/etc/dcgm-exporter/default-counters.csv}"
+
--- a/tests/e2e/Makefile
+++ b/tests/e2e/Makefile
@@ -19,31 +19,27 @@ IMAGE_REPOSITORY ?= "nvcr.io/nvidia/k8s/dcgm-exporter"
 IMAGE_TAG ?= "3.3.5-3.4.0-ubuntu22.04"
 KUBECONFIG ?= "~/.kube/config"
 
-.PHONY: e2e-test
-e2e-test:
+define TEST_CMD
 	@if [ -z ${KUBECONFIG} ]; then \
 		echo "[ERR] KUBECONFIG is missing, must be set"; \
 		exit 1; \
 	fi
-	$(GO_CMD) test -failfast --tags=e2e -v . -args \
+	$(GO_CMD) test --tags=e2e -v . \
+		-args \
 		-kubeconfig=$(KUBECONFIG) \
 		-chart="$(CHART)" \
 		-namespace=$(NAMESPACE) \
 		-image-repository=$(IMAGE_REPOSITORY) \
 		-image-tag=$(IMAGE_TAG)
+endef
+
+.PHONY: e2e-test
+e2e-test:
+	@$(TEST_CMD)
 
 
 .PHONY: e2e-test-no-profiling
 e2e-test-no-profiling:
-	@if [ -z ${KUBECONFIG} ]; then \
-		echo "[ERR] KUBECONFIG is missing, must be set"; \
-		exit 1; \
-	fi
-	$(GO_CMD) test -failfast --tags=e2e -v . -args \
-		-kubeconfig=$(KUBECONFIG) \
-		-chart="$(CHART)" \
-		-namespace=$(NAMESPACE) \
-		-image-repository=$(IMAGE_REPOSITORY) \
-		-image-tag=$(IMAGE_TAG) \
+	@$(TEST_CMD) \
 		-arguments="{-f=/etc/dcgm-exporter/default-counters.csv}"
 
--- a/tests/e2e/e2e_suite_test.go
+++ b/tests/e2e/e2e_suite_test.go
@@ -64,6 +64,7 @@ type suiteConfig struct {
 	chart           string
 	imageRepository string
 	imageTag        string
+	arguments       string
 }
 
 type Suite struct {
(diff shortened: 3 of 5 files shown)


.vscode/launch.json
```diff
@@ -4,6 +4,16 @@
...
+            "name": "Go: Launch e2e",
+            "type": "go",
+            "request": "launch",
+            "mode": "test",
+            "program": "${workspaceFolder}/tests/e2e",
+            "args": ["-kubeconfig","~/.kube/config", "-chart","./../../deployment/","-image-repository","nvidia/dcgm-exporter","-arguments","{-f=/etc/dcgm-exporter/dcp-metrics-included.csv,--enable-dcgm-log=true,--dcgm-log-level=ERROR}"],
+            "env": {},
+            "buildFlags": "-tags=e2e"
+        },
         {
```
nvvfedorov, line 17: Added example, how to run and debug e2e tests in the VS.CODE

tests/e2e/Makefile
```diff
@@ -25,9 +25,25 @@ e2e-test:
 		echo "[ERR] KUBECONFIG is missing, must be set"; \
 		exit 1; \
 	fi
-	$(GO_CMD) test --tags=e2e -v . -args \
+	$(GO_CMD) test -failfast --tags=e2e -v . -args \
 		-kubeconfig=$(KUBECONFIG) \
```
nvvfedorov, line 29: The failfast flag is required to ensure that the test suite fails after 1st failed test.

tests/e2e/Makefile
```diff
@@ -25,9 +25,25 @@ e2e-test:
...
 		-chart="$(CHART)" \
 		-namespace=$(NAMESPACE) \
 		-image-repository=$(IMAGE_REPOSITORY) \
 		-image-tag=$(IMAGE_TAG)
+
+
+.PHONY: e2e-test-no-profiling
+e2e-test-no-profiling:
+	@if [ -z ${KUBECONFIG} ]; then \
```
nvvfedorov, line 38: Added option to run e2e tests on GPU without profiling support.

tests/e2e/main_test.go
```diff
@@ -66,6 +66,11 @@ func TestMain(m *testing.M) {
 		"",
 		"DCGM-exporter image tag to use")
 
+	flag.StringVar(&suiteCfg.arguments,
+		"arguments",
+		"",
+		"DCGM-exporter command line")
+
 	flag.Parse()
```
nvvfedorov, line 74: Arguments allow to pass a custom set of arguments to the dcgm-exporter helm chart for deployment.

tests/e2e/e2e_suite_test.go
```diff
@@ -190,11 +195,9 @@ func (s *Suite) TearDownSuite() {
 	s.T().Log("Starting tear down E2E test setup...")
 
 	if s.workloadPod != nil {
-		s.T().Logf("Starting delete of the workload pod: %s...", s.workloadPod.Name)
```
nvvfedorov, line 193: The issue, with `s.T().Logf` that messages will appear only after test execution when golang test framework flushes the output buffer. It doesn't make sense to show the message after the fact that the workload pod was deleted.

tests/e2e/e2e_suite_test.go
```diff
@@ -226,8 +228,9 @@ func (s *Suite) TearDownSuite() {
 func (s *Suite) TestDCGMExporter() {
 	s.DCGMExporterPrechecks()
 
-	s.T().Run("Create workload pod", func(t *testing.T) {
-		t.Log("Creating a workload pod...")
```
nvvfedorov, line 230: Refactoring, to use the testify Suite framework to run sub-tests, which makes much easier work with the tree of tests.

tests/e2e/Makefile
```diff
@@ -25,9 +25,25 @@ e2e-test:
...
+		exit 1; \
+	fi
+	$(GO_CMD) test -failfast --tags=e2e -v . -args \
+		-kubeconfig=$(KUBECONFIG) \
+		-chart="$(CHART)" \
+		-namespace=$(NAMESPACE) \
+		-image-repository=$(IMAGE_REPOSITORY) \
+		-image-tag=$(IMAGE_TAG) \
+		-arguments="{-f=/etc/dcgm-exporter/default-counters.csv}"
+
```
rohit-arora-dev, line 49: To avoid duplication, it can be a function:
```
define TEST_CMD
	@if [ -z $${KUBECONFIG} ]; then \
		echo "[ERR] KUBECONFIG is missing, must be set"; \
		exit 1; \
	fi; \
	$(GO_CMD) test -failfast --tags=e2e -v . -args \
		-kubeconfig=$${KUBECONFIG} \
		-chart="$${CHART}" \
		-namespace=$${NAMESPACE} \
		-image-repository=$${IMAGE_REPOSITORY} \
		-image-tag=$${IMAGE_TAG}
endef
```

then use:
```
e2e-test-no-profiling:
	@$(TEST_CMD) \
		-arguments="{-f=/etc/dcgm-exporter/default-counters.csv}"
```

tests/e2e/main_test.go
```diff
@@ -66,6 +66,11 @@ func TestMain(m *testing.M) {
 		"",
 		"DCGM-exporter image tag to use")
 
+	flag.StringVar(&suiteCfg.arguments,
+		"arguments",
+		"",
+		"DCGM-exporter command line")
```
rohit-arora-dev, line 72: Would be nice to given an example in the comments, how you expect a sample argument to look like esp if you have 2-3 arguments.

tests/e2e/internal/framework/kube.go
```diff
@@ -92,6 +92,12 @@ func (c *KubeClient) CheckPodCondition(ctx context.Context,
 		}
 	}
 
+	for _, c := range pod.Status.ContainerStatuses {
+		if c.State.Waiting != nil && c.State.Waiting.Reason == "CrashLoopBackOff" {
+			return false, fmt.Errorf("pod %s in namespace %s is in CrashLoopBackOff", pod.Name, pod.Namespace)
+		}
+	}
+
```
rohit-arora-dev, line 100: Not clear why you also need to query containr status.
Also, this breaks the single responsibility rule now this function is doing two different things: 1. Checking pod status, 2. checking underlying container statuses.

tests/e2e/Makefile
nvvfedorov: @rohit-arora-dev , Thank you for the advise.

tests/e2e/e2e_suite_test.go
```diff
@@ -58,218 +53,285 @@ const (
...
+					dcgmExpPod.Name,
+					func(namespace, podName string, status corev1.PodStatus) (bool, error) {
+						for _, c := range status.Conditions {
+							if c.Type != corev1.PodReady {
+								continue
+							}
+							if c.Status == corev1.ConditionTrue {
+								return true, nil
+							}
+						}
```
nvvfedorov, line 247: @rohit-arora-dev, We must wait for the status: PodReady or "CrashLoopBackOff." The check for the  "CrashLoopBackOff" is necessary to interrupt the waiting if the pod crashes. If the pod crashes, waiting until it is up and running doesn't make sense. With the implemented logic, I reduced the waiting time from 15 minutes to a few seconds in case we see that the pod crashes.

.vscode/launch.json
```diff
@@ -4,6 +4,16 @@
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
+        {
+            "name": "Go: Launch e2e",
+            "type": "go",
+            "request": "launch",
+            "mode": "test",
+            "program": "${workspaceFolder}/tests/e2e",
+            "args": ["-kubeconfig","~/.kube/config", "-chart","./../../deployment/","-image-repository","nvidia/dcgm-exporter","-arguments","{-f=/etc/dcgm-exporter/dcp-metrics-included.csv,--enable-dcgm-log=true,--dcgm-log-level=ERROR}"],
```
rohit-arora-dev, line 13: nit:
Either in this change or the follow-up, it would be nice to list them in sepeate line for improved readability. Just like line 24.

tests/e2e/main_test.go
```diff
@@ -66,10 +66,15 @@ func TestMain(m *testing.M) {
...
+	flag.StringVar(&suiteCfg.arguments,
+		"arguments",
+		"",
+		`DCGM-exporter command line parameters. Example: -arguments={-f=/etc/dcgm-exporter/default-counters.csv}`)
+
 	flag.Parse()
 	os.Exit(m.Run())
 }
-
+git
```
rohit-arora-dev, line 77: may be an accident.

rohit-arora-dev (approved):
A minor accidental editting, otherwise looking good.

/LGTM

tests/e2e/main_test.go
nvvfedorov: Oops, @rohit-arora-dev , thank you for finding it. I have already pushed the change.

rohit-arora-dev (approved):

PR #298, opened 2024-03-20
Fix CPU Core Group Accounting

rohit-arora-dev:
This fixes a minor bug when creating Groups to watch CPU Cores. This ensures a CPU core that is not being watched is not accounted towards the collector group capacity.


glowkey (approved):
LGTM

PR #332, opened 2024-05-23
Update Makefile

jjziets:
added DIST_DIR       := cmd/dcgm-exporter


rvatkar:
It introduces a dead variable. As-is, it's an incomplete change with no functional impact.

PR #346, opened 2024-06-25
fix: correct metric help text

pintohutch:
A few of the profiling metric gauges claimed a % unit in the help text of the sample CSVs. However, they are reported as a ratio [0,1].

This change removes the percentage unit described for correctness.


glowkey (approved):

pintohutch:
Thanks @glowkey - I just updated the commit to be properly signed per the PR workflow requirements.

Issue #353, opened 2024-07-05
Duplicated / wrong metrics if using MIG, Grafana dashboard showing wrong duplicated  / false values


frittentheke:
### What is the version?

3.3.5-3.4.1

### What happened?

When activating MIG we saw duplicated and plain wrong metrics in the provided Grafana dashboard (https://github.com/NVIDIA/dcgm-exporter/tree/main/grafana).

The issue seems to be two-fold, with Grafana as well as the raw metrics themselves:

1. Firstly the dashboard (legends, ...) and PromQL queries used to fetch metrics do not take MIG into account. So metrics returning MIG subdevices (`GPU_I_ID`) are not considered.
2. Even if the queries were updated via e.g some aggregations like `max()`, `avg() `or `sum()` to avoid duplication, there are some metrics reported back per `GPU_I_ID`, that do not have this granularity. See me comment https://github.com/NVIDIA/dcgm-exporter/issues/257#issuecomment-2210537130. So if the power draw is not measured per `GPU_I_ID` you cannot return it individually as you would be returning false values.

### What did you expect to happen?

Provided MIG and other ways of partitioning GPUs (vGPU, time-slicing, ...) is quite common, I'd expect the exporter and the provided dashboard to take those into account.

Metrics that are available per-subdevice should be returned, if they are just duplicates of each other, they should be dropped and only returned per "main" GPU.

### What is the GPU model?

H100s, using different MIG profiles and whole GPUs

### What is the environment?

Kubernetes

### How did you deploy the dcgm-exporter and what is the configuration?

Kubernetes with GPU-Operator

### How to reproduce the issue?

Enable MIG on a GPU and look at the dashboard.

nvvfedorov:
@frittentheke, Thank you for reporting about the issue. Am I right that the main request is the following: Metrics that are available per-subdevice should be returned; if they are just duplicates of each other, they should be dropped and only returned per "main" GPU.

frittentheke:
Thanks @nvvfedorov  for your fast response!

> @frittentheke, Thank you for reporting about the issue. Am I right that the main request is the following: Metrics that are available per-subdevice should be returned; if they are just duplicates of each other, they should be dropped and only returned per "main" GPU.

1. Yes. Please also see my PR in which I (had to) applied aggregations like max() to work around this for the dashboard.
[...]

PR #240, opened 2024-01-29
Fix grafana dashboard cannot display properly in vGPU cluster

Levi080513:
Fix https://github.com/NVIDIA/dcgm-exporter/issues/236

### Test

Create a k8s cluster with vGPU configured on one node.
```
kc get node hw-sks-test-vgpu-vgpunode-8jwnn -oyaml | yq '.status.allocatable'
cpu: "8"
ephemeral-storage: "57976119610"
hugepages-2Mi: "0"
memory: 15968092Ki
nvidia.com/gpu: "1"
pods: "110"

kc exec -ti nvidia-driver-daemonset-4.18.0-477.27.1.el8.8-rocky8.8-x7qj8 -n sks-system-nvidia-gpu -- nvidia-smi
Mon Jan 29 10:20:21 2024
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 525.105.17   Driver Version: 525.105.17   CUDA Version: 12.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  GRID V100-4C        On   | 00000000:00:0A.0 Off |                    0 |
| N/A   N/A    P0    N/A /  N/A |     12MiB /  4096MiB |      0%      Default |
|                               |                      |             Disabled |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|    0   N/A  N/A    243152      C   /app/gpu_burn                      12MiB |
+-----------------------------------------------------------------------------+
```

Before fixing
image

After fixing
image

image

Issue #236, opened 2024-01-22
NVIDIA DCGM Exporter Dashboard does not work in vGPU cluster


Levi080513:
Currently we use `DCGM_FI_DEV_GPU_TEMP` to obtain the instance/GPU list, but this metrics is not collected in vGPU clusters. This will prevent the dashboard from displaying properly.

https://github.com/NVIDIA/dcgm-exporter/blob/30d4ddcae9c7153c31dd35301aa4a1f3b90a2096/grafana/dcgm-exporter-dashboard.json#L784

https://github.com/NVIDIA/dcgm-exporter/blob/30d4ddcae9c7153c31dd35301aa4a1f3b90a2096/grafana/dcgm-exporter-dashboard.json#L761

nvvfedorov:
Can you try to use other metrics available on your vGPU?

Levi080513:
`DCGM_FI_DEV_GPU_UTIL` metrics is work well.

Levi080513:
Can I submit a PR to fix it?

nvvfedorov:
@Levi080513 , sure you can submit PRs; we appreciate community contribution.

troublesprouter:
You may want to try this dashboard out to view GPU utilization: https://github.com/troublesprouter/gpu-util-dashboard
I haven't tried it with vGPUs but it may work.

PR #355, opened 2024-07-08
[dashboard] Rework dashboard (MIG support, Grafana deprecations, Hostname)

frittentheke:
Running into various issues with the dashboard (see  #353) I started reworking the existing board.
This PR combines all my cleanups and fixes. It also includes the changes of PR https://github.com/NVIDIA/dcgm-exporter/pull/240 by @Levi080513

* Use PromQL aggregations to take MIG subdevices into account (see #353)
* Update all panels to use Timeseries panels (instead of deprecated Graph)
* Switch from instance to Hostname to select individual systems to avoid
  duplicated timeseries for Kubernetes daemonsets and their Pod names
* Use DCGM_FI_DEV_GPU_UTIL instead of DCGM_FI_DEV_GPU_TEMP to also cover vGPU (PR #240)
* Use DCGM_FI_PROF_GR_ENGINE_ACTIVE to determine utilization to cover MIG (and vGPU)

Fixes: #353, #236


frittentheke:
^^ @rohit-arora-dev @glowkey ?

frittentheke:
> I just took a moment to test these changes on an 8 GPU system with MIG enabled and unfortunately the panels were empty. I'm far from a Grafana expert so it's hard for me to know what was going wrong. I did confirm that without the changes the panels displayed the expected data.

Some more details on which graph and with which PromQL query doesn't work would be great.

kristiangronas:
For the GPU Util i think the old way may be better, the prof module is proprietary and only supported on a small amount of gpus / configurations

The other changes seem great though

https://github.com/NVIDIA/dcgm-exporter/issues/380 it looks like most of the prom metrics are not really reliable enough to be used as the sole source of labels in all the different situations the exporter is used

Issue #314, opened 2024-04-12
The pod for a given GPU in k8s mode cannot be captured


rokkiter:
### What happened?

Unable to collect GPU metrics for relevant pods when using passthrough mode.
For example, dcgm-exporter does not collect metrics when a VM created with kubevirt mounts a GPU in passthrough mode.

kubevirt vmi yaml for mounting GPUs
```
spec:
  domain:
    devices:
      ...
      gpus:
      - deviceName: nvidia.com/GPUDeviceName
        name: gpu1
```

The resource of the kubevirt launcher pod that needs to be monitored.

```
resources:
  ...
  requests:
    ...
    nvidia.com/GPUDeviceName: "1"
```

In the following code, we can find the rules are resourceName == nvidiaResourceName or strings.HasPrefix(resourceName, nvidiaMigResourcePrefix) The nvidiaResourceName is "nvidia.com/gpu"
This filters the mounting of specific devices.

https://github.com/NVIDIA/dcgm-exporter/blob/main/pkg/dcgmexporter/kubernetes.go#L142
```
func (p *PodMapper) toDeviceToPod(
	devicePods *podresourcesapi.ListPodResourcesResponse, sysInfo SystemInfo,
) map[string]PodInfo {
	deviceToPodMap := make(map[string]PodInfo)

	for _, pod := range devicePods.GetPodResources() {
		for _, container := range pod.GetContainers() {
			for _, device := range container.GetDevices() {

				resourceName := device.GetResourceName()
				if resourceName != nvidiaResourceName {
					// Mig resources appear differently than GPU resources
					if !strings.HasPrefix(resourceName, nvidiaMigResourcePrefix) {
						continue
					}
				}
				...
			}
		}
	}

	return deviceToPodMap
}
```

### What did you expect to happen?

GPU metrics can be collected when mounting a GPU card using kubevirt passthrough mode

### What is the environment?

pod

### How did you deploy the dcgm-exporter and what is the configuration?

GPU Operator

### How can we reproduce the issue?

Mounting a GPU card using kubevirt passthrough mode.

### What is the version?

Latest

### Anything else we need to know?

Some discussions in the kubevirt community.
https://github.com/kubevirt/kubevirt/issues/11660

nvvfedorov:
[...]

PR #359, opened 2024-07-12
Make nvidia resource names configurable

lx1036:
fix issue https://github.com/NVIDIA/dcgm-exporter/issues/314 , to make nvidia resource names configurable.

It can solve the problem that nvidia/k8s-device-plugin advertise specified GPU type like nvidia.com/a100 or nvidia.com/a10 and so on, not nvidia.com/gpu.


nvvfedorov (approved):
LGTM

lx1036:
@nvvfedorov have added a GPG signature to my commit. thanks for review.

PR #366, opened 2024-07-24
DCGM-Exporter release version 3.3.7-3.5.0

glowkey:


docker/Dockerfile.ubuntu22.04
```diff
@@ -1,5 +1,5 @@
-FROM nvcr.io/nvidia/cuda:12.4.1-base-ubuntu22.04 AS builder
-ARG GOLANG_VERSION=1.21.5
+FROM nvcr.io/nvidia/cuda:12.5.1-base-ubuntu22.04 AS builder
+ARG GOLANG_VERSION=1.22.4
```
rohit-arora-dev, line 2: Do we need to update Golang to 1.22.5 here as well, or I guess the Makefile overrides it.

docker/Dockerfile.ubi9
```diff
@@ -1,5 +1,5 @@
-FROM nvcr.io/nvidia/cuda:12.4.1-base-ubi9 AS builder
-ARG GOLANG_VERSION
+FROM nvcr.io/nvidia/cuda:12.5.1-base-ubi9 AS builder
+ARG GOLANG_VERSION=1.22.4
```
rohit-arora-dev, line 2: Do we need to update Golang to 1.22.5 here as well, or I guess the Makefile overrides it.

go.mod
```diff
@@ -1,16 +1,18 @@
 module github.com/NVIDIA/dcgm-exporter
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.4
```
rohit-arora-dev, line 5: Do we need to update Golang to 1.22.5 here as well?

rohit-arora-dev (commented):
Just a common comment related to Go version, making sure if it is what is expected.

docker/Dockerfile.ubuntu22.04
rohit-arora-dev: Discussed offline, marking resolved.

rohit-arora-dev (approved):
/LGTM

Issue #381, opened 2024-08-30
Error with "make binary" operation in local development


kschoi93:
### What is the version?

3.3.7-3.5.0

### What happened?

this repository needs to edit its ".gitignore" for local development

develop dcgm exporter -> git commit & push -> git pull from another docker -> dcgm bianry or install -> error

---

root@dcgm-exporter-vzjgc:/custom-dcgm-exporter# make binary
...
go.uber.org/mock/mockgen
Updating DCGM version in files from 3.3.6 to 3.3.7...
Updating DCGM Exporter version in files from 3.4.2 to 3.5.0...
cd cmd/dcgm-exporter; go build -ldflags "-X main.BuildVersion=3.3.7-3.5.0"
/bin/sh: 1: cd: can't cd to cmd/dcgm-exporter
no Go files in /surromind-dcgm-exporter
make: *** [Makefile:36: binary] Error 1

---

root@dcgm-exporter-vzjgc:/surromind-dcgm-exporter# ll
total 208
drwxr-xr-x 16 root root  4096 Aug 30 02:13 ./
drwxr-xr-x  1 root root  4096 Aug 30 02:07 ../
drwxr-xr-x  2 root root  4096 Aug 30 02:07 .devcontainer/
drwxr-xr-x  8 root root  4096 Aug 30 02:14 .git/
drwxr-xr-x  4 root root  4096 Aug 30 02:07 .github/
-rw-r--r--  1 root root  5418 Aug 30 02:13 .gitignore
-rw-r--r--  1 root root     0 Aug 30 02:07 .gitmodules
-rw-r--r--  1 root root   283 Aug 30 02:07 .golangci.yml
drwxr-xr-x  2 root root  4096 Aug 30 02:07 .vscode/
-rw-r--r--  1 root root  2242 Aug 30 02:07 CONTRIBUTING.md
-rw-r--r--  1 root root  1543 Aug 30 02:07 Jenkinsfile
-rw-r--r--  1 root root 11358 Aug 30 02:07 LICENSE
-rw-r--r--  1 root root  4525 Aug 30 02:12 Makefile
-rw-r--r--  1 root root  8545 Aug 30 02:12 README.md
-rw-r--r--  1 root root  1004 Aug 30 02:07 RELEASE.md
-rw-r--r--  1 root root  2082 Aug 30 02:12 dcgm-exporter.yaml
drwxr-xr-x  3 root root  4096 Aug 30 02:12 deployment/
[...]

PR #382, opened 2024-08-30
fix: edit gitignore and require dir & file

kschoi93:
Set the "/cmd/dcgm-exporter" folder, which must be included, in gitignore.

Please refer to the issue contents
#381

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.swp
 *.swo
 dcgm-exporter
+!/cmd/dcgm-exporter
 !etc/
 !deployment/
 .env


glowkey:
I could not reproduce #381 on a fresh OS instance. We do not plan to merge this at it seems very specific to your environment.

kschoi93:
@glowkey.
I'm guessing you've probably been working on a forked repository. 
If you did "git clone" -> "another git repo init" -> "git add .", I can confirm that the issue is caused by .gitignore.


cmd_not_added

kschoi93:
You may ask why I clone the original git repository of dcgm exporter instead of doing git fork & clone. 
If you do git clone, there is an example in the README. 
If you use that example to develop and do git push to the git repository you created, you may get the same error I wrote about.

building from source

kschoi93:
@glowkey  Please confirm again.

glowkey:
No, we do not plan on making this change. Please fork your own repository if you're attempting to make changes.

kschoi93:
Since you can't reproduce the problem, I personally proved that it occurs in certain cases and that it is not limited to my specific environment. 
The problem is that you are using gitignore incorrectly in this project, and I asked you to fix it. 
I don't understand why you don't fix it when you can see that your gitignore is incorrect just by looking at the image, not the text above.

PR #391, opened 2024-09-21
Service monitor API value configurable

dtzar:
Allows the service monitor API to be configurable for other providers


nvvfedorov (approved):
LGTM

PR #426, opened 2024-11-26
Fix empty clusterIP when no clusterIP is defined in values

nebula-it:
```
clusterIP:
```
Should only show up in manifest if `.Values.service.clusterIP` is defined. Otherwise it shows an empty value, which causes argoCD to report app as out of sync because of expects `spec.clusterIP` to have the auto assigned IP but manifest states it should be null value.

### DIFF
--- a/deployment/templates/service.yaml
+++ b/deployment/templates/service.yaml
@@ -27,7 +27,7 @@ metadata:
   {{- end }}
 spec:
   type: {{ .Values.service.type }}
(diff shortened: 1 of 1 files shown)


glowkey (approved):

c0deaddict:
Thanks for fixing this! I don't think there is a release yet that contains this fix? Can a maintainer create a release please? :pray:

nebula-it:
Yeah there isn't. Maybe @glowkey can help us with that

Issue #307, opened 2024-04-05
Per pod metrics not exposed with time-slicing enabled


ThisIsQasim:
### What is the version?

3.3.5-3.4.1

### What happened?

Metrics like `DCGM_FI_PROF_GR_ENGINE_ACTIVE` are only exposed for one single pod even though there are multiple pods that use the same GPU

### What did you expect to happen?

Metrics for the all the pods should be exposed

### What is the GPU model?

Tesla T4

### What is the environment?

GKE

### How to reproduce the issue?

- Enable time-slicing using device plugin
- Deploy DCGM and dcgm-exporter
- Deploy app that uses GPU
- Check metrics

### Anything else we need to know?

From the debug log
```
time="2024-04-05T13:49:04Z" level=debug msg="Device to pod mapping: map[nvidia0:{Name:gpu-pod-c69f6664f-vkkcb Namespace:default Container:extractor} nvidia0/vgpu0:{Name:gpu-pod-c69f6664f-vkkcb Namespace:default Container:extractor} nvidia0/vgpu1:{Name:gpu-pod-c69f6664f-2v922 Namespace:default Container:extractor} nvidia0/vgpu2:{Name:gpu-pod-c69f6664f-wrcxw Namespace:default Container:extractor} nvidia0/vgpu3:{Name:gpu-pod-c69f6664f-ffs8r Namespace:default Container:extractor}]"
```

ThisIsQasim:
This appears to have been reported repeatedly #151 #201  #222

nvvfedorov:
The performance metrics require exclusive access to GPU hardware with Turing architecture. If another pod tries to read performance metrics, the DCGM exporter cannot read performance metrics.

ThisIsQasim:
there is only pod per node trying to read the metrics but multiple pods using the same GPU. The issue is that dcgm exporter should report metrics for all the pods using the GPU.

ThisIsQasim:
Sure. A single GPU is advertised as multiple using the nvidia device plugin
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: time-slicing-config
data:
  any: |-
    version: v1
    flags:
      migStrategy: none
    sharing:
      timeSlicing:
        renameByDefault: false
        failRequestsGreaterThanOne: false
        resources:
          - name: nvidia.com/gpu
            replicas: 4
```
and then GPUs are requested with the regular resource requests
```
esources:
  requests:
    cpu: 3600m
  limits:
[...]

PR #432, opened 2024-12-12
feat: add support for gpu sharing metrics in k8s

pintohutch:
We add support for capturing separate metrics when running dcgm-exporter in K8s clusters that have GPU sharing enabled, including time-sharing and MPS. This should now support GPU sharing on MIG devices as well.

We ensure this is supported for both the NVIDIA and GKE device plugins, respectively at:
* https://github.com/NVIDIA/k8s-device-plugin
* https://github.com/GoogleCloudPlatform/container-engine-accelerators

First, we make a small fix to the Kubernetes PodMapper tranform processor. Specifically we update the regular expression used in building the device mapping to properly capture pod attributes in both MIG and MIG-with-sharing GPUs in GKE.

The bulk of the change is guarded by a new configuration parameter, which can be passed in as a flag `--kubernetes-virtual-gpus` or as an environment variable `KUBERNETES_VIRTUAL_GPUS`. If set, the Kubernetes PodMapper tranform processor uses a different mechanism to build the device mapping, which creates a copy of the metric for every shared (i.e. virtual) GPU exposed by the device plugin. To disambiguate the generated timeseries, it adds a new label "vgpu" set to the detected shared GPU replica.

This also fixes an issue where pod attributes are not guaranteed to be consistently associated with the same metric. If the podresources API does not consistently return the device-ids in the same order between calls, the device-to-pod association in the map can change between scrapes due to an overwrite that happens in the Process loop.

Ultimately, we may wish to make this the default behavior. However, guarding it behind a flag:
1. Mitigates any risk of the change in case of bugs
2. Given the feature adds a new label, it is possible PromQL queries performing deaggregation using, e.g.  `without` or `ignore` functions, may break existing dashboards and alerts. Allowing users to opt-in via a flag ensures backwards compatibility in these scenarios.

Finally, we update the unit tests to ensure thorough coverage for the changes.

### DIFF
--- a/pkg/cmd/app.go
+++ b/pkg/cmd/app.go
@@ -75,6 +75,7 @@ const (
 	CLIPodResourcesKubeletSocket  = "pod-resources-kubelet-socket"
 	CLIHPCJobMappingDir           = "hpc-job-mapping-dir"
 	CLINvidiaResourceNames        = "nvidia-resource-names"
+	CLIKubernetesVirtualGPUs      = "kubernetes-virtual-gpus"
 )
 
 func NewApp(buildVersion ...string) *cli.App {
@@ -243,6 +244,12 @@ func NewApp(buildVersion ...string) *cli.App {
 			Value:   cli.NewStringSlice(),
 			Usage:   "Nvidia resource names for specified GPU type like nvidia.com/a100, nvidia.com/a10.",
 			EnvVars: []string{"NVIDIA_RESOURCE_NAMES"},
+		},
+		&cli.BoolFlag{
+			Name:    CLIKubernetesVirtualGPUs,
+			Value:   false,
+			Usage:   "Capture metrics associated with virtual GPUs exposed by Kubernetes device plugins when using GPU sharing strategies, e.g. time-sharing or MPS.",
+			EnvVars: []string{"KUBERNETES_VIRTUAL_GPUS"},
 		},
 	}
 
@@ -639,5 +646,6 @@ func contextToConfig(c *cli.Context) (*dcgmexporter.Config, error) {
 		PodResourcesKubeletSocket:  c.String(CLIPodResourcesKubeletSocket),
 		HPCJobMappingDir:           c.String(CLIHPCJobMappingDir),
 		NvidiaResourceNames:        c.StringSlice(CLINvidiaResourceNames),
+		KubernetesVirtualGPUs:      c.Bool(CLIKubernetesVirtualGPUs),
 	}, nil
 }
--- a/pkg/dcgmexporter/config.go
+++ b/pkg/dcgmexporter/config.go
@@ -59,4 +59,5 @@ type Config struct {
 	PodResourcesKubeletSocket  string
 	HPCJobMappingDir           string
 	NvidiaResourceNames        []string
+	KubernetesVirtualGPUs      bool
 }
--- a/pkg/dcgmexporter/kubernetes.go
+++ b/pkg/dcgmexporter/kubernetes.go
@@ -36,7 +36,8 @@ import (
 var (
 	connectionTimeout = 10 * time.Second
 
-	gkeMigDeviceIDRegex            = regexp.MustCompile(`^nvidia([0-9]+)/gi([0-9]+)$`)
+	// Allow for MIG devices with or without GPU sharing to match.
+	gkeMigDeviceIDRegex            = regexp.MustCompile(`^nvidia([0-9]+)/gi([0-9]+)(/vgpu[0-9]+)?$$`)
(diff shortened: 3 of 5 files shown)


pintohutch:
Fixed #307

glowkey:
Thank you for the contribution! We will test and review the PR in the coming weeks.

Please be aware that we are on the verge of releasing a new 4.0 version of DCGM-Exporter so this PR will likely need to be updated after that happens.

pintohutch:
Thanks @glowkey - I actually may factor out the smaller bugfix in a separate PR:

> First, we make a small fix to the Kubernetes PodMapper tranform processor. Specifically we update the regular expression used in building the device mapping to properly capture pod attributes in both MIG and MIG-with-sharing GPUs in GKE.

That way, this change just focuses on the new `KUBERNETES_VIRTUAL_GPUS` addition.

pintohutch:
Moved the bugfix fix over to https://github.com/NVIDIA/dcgm-exporter/pull/433

glowkey (approved):

arthas3014:
I started several Pods on the A100 using mps, and running dcgm-exporter with env: KUBERNETES_VIRTUAL_GPUS: true,although each pod can be bound to an metric with label vgpu:xx,but all the pod metrics have the same value in the time series.Each pod should use a different case of gpu, otherwise these same value of pod indicators would not be meaningful?

pintohutch:
@arthas3014 - from my understanding DCGM metrics are only recorded per unique GPU UUID, so what you're describing is WAI.

Sharing strategies leveraging K8s device-plugins involve exposing a physical GPU as multiple "virtual" GPUs to the kubelet so multiple pods can be scheduled to share the hardware.

Without `KUBERNETES_VIRTUAL_GPUS`, the dcgm-exporter would associate a pod to a DCGM metric inconsistently, as explained in the PR description. i.e. If `pod-1` and `pod-2` were sharing a GPU, then on a given Prometheus scrape, `pod-1` would be associated with the metric, while the next scrape could associate it with `pod-2`. This change ensures a consistent mapping on every scrape by adding the new `vgpu` label. In theory, each timeseries should contain a sample on every scrape now.

IMO this improves the performance of queries asking "what did the GPU look like when `pod-1` was running?". But AFAIK DCGM is not able to disambiguate what processes are using the GPU at a given time. @glowkey to keep me honest 🙂

kev-iotairx:
@pintohutch @glowkey Is this part of the latest dcgm-exporters ? We tried adding the environment variable `KUBERNETES_VIRTUAL_GPUS` but we are not seeing any data being exposed. 

Here's a snippet from our `dcgm-exporter` pod: 

```
Pod Template:
  Labels:           app.kubernetes.io/name=pebble-falcon-dcgm-exporter
  Service Account:  pebble-falcon-dcgm-account
  Containers:
   pebble-falcon-dcgm:
    Image:      nvcr.io/nvidia/k8s/dcgm-exporter:4.2.3-4.1.3-ubuntu22.04
    Port:       9400/TCP
    Host Port:  0/TCP
    Args:
      -r
      $(NODE_IP):5555
      -f
      /etc/dcgm-exporter/default-counters.csv
```

When we remove the `KUBERNETES_VIRTUAL_GPUS` env variable from the pod, we see the expected metrics at the aggregate level but unavailable at the pod-level. Can you share the latest on this?

kev-iotairx:
How do I interpret these metrics in the context of MPS, where I have multiple pods running and sharing resources concurrently? 

For instance, the 2 workloads I have running have completely distinct profiles. But, the metrics for them are exactly the same. 

```
# HELP DCGM_FI_DEV_FB_USED Framebuffer memory used (in MiB).
# TYPE DCGM_FI_DEV_FB_USED gauge
DCGM_FI_DEV_FB_USED{gpu="0",UUID="GPU-3bc239e1-4cc8-97e4-e817-5a36dfd2885e",pci_bus_id="00000000:00:05.0",device="nvidia0",modelName="NVIDIA A100-SXM4-80GB",Hostname="gke-gpu-share-test-a100-inference-mps-b3fc3d0e-3bpg",container="gpu-burn-container-1",namespace="default",pod="gpu-burn-deployment-1-85dc8ff699-clvh5",vgpu="1"} 15866

DCGM_FI_DEV_FB_USED{gpu="0",UUID="GPU-3bc239e1-4cc8-97e4-e817-5a36dfd2885e",pci_bus_id="00000000:00:05.0",device="nvidia0",modelName="NVIDIA A100-SXM4-80GB",Hostname="gke-gpu-share-test-a100-inference-mps-b3fc3d0e-3bpg",container="vllm",namespace="vllm-inference",pod="vllm-medgemma-4b-it-d599b4c99-f95nk",vgpu="3"} 15866
```

PR #433, opened 2024-12-14
fix: enable pod attributes for gpu-sharing in gke

pintohutch:
We make a small fix to the Kubernetes PodMapper tranform processor.

Specifically we update the regular expression used in building the device mapping to properly capture pod attributes in both MIG and [MIG-with-sharing](https://cloud.google.com/kubernetes-engine/docs/how-to/timesharing-gpus#use-with-multi-instance-gpu) GPUs in GKE.

### DIFF
--- a/internal/pkg/transformation/kubernetes.go
+++ b/internal/pkg/transformation/kubernetes.go
@@ -41,7 +41,8 @@ import (
 var (
 	connectionTimeout = 10 * time.Second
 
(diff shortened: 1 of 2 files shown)


pintohutch:
FYI @glowkey - I rebased against `main` to incorporate the DCGM 4 changes from #437

glowkey (approved):
LGTM

PR #435, opened 2024-12-30
feat: add log formatting

pintohutch:
We introduce log formatting to support both text and json formats. This can be configured through a CLI flag "log-format" or env var "DCGM_EXPORTER_LOG_FORMAT".

Notably, the log format will capture any DCGM logging that is configured.

We also move the automaxprocs call from an unused import to be explicitly called when starting the dcgm exporter to ensure its logging format honors the configuration.

### DIFF
--- a/cmd/dcgm-exporter/main.go
+++ b/cmd/dcgm-exporter/main.go
@@ -20,8 +20,6 @@ import (
 	"log/slog"
 	"os"
 
-	_ "go.uber.org/automaxprocs"
-
 	"github.com/NVIDIA/dcgm-exporter/pkg/cmd"
 )
 
--- a/pkg/cmd/app.go
+++ b/pkg/cmd/app.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/NVIDIA/go-dcgm/pkg/dcgm"
 	"github.com/urfave/cli/v2"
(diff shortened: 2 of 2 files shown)


pintohutch:
Friendly ping @glowkey - is this something that could be useful?

glowkey:
Yes, can it be adjusted to handle the user specified log-level

glowkey:
Oh, and thank you for the submission!

glowkey:
In my testing this change was incompatible with the Capture mechanism and I wasn't seeing any output in either format. When I disabled Capture then I was able to test the change and toggle between text and json. My test was simply running the binary without any args. How did you test the change?

pintohutch:
Hrm - thanks for catching. I believe this was working before the 4.X release, but I see the same thing now.

It may be more involved than I originally thought. I'm less confident how to untangle the Capture mechanism...

pintohutch:
Ah - I think it just needed to route to `os.Stderr` - patched that and tested it. PTAL.

pintohutch:
Did a few other fixes to handle debug logging and automaxprocs properly.

pintohutch:
Friendly ping @glowkey

glowkey (approved):
lgtm

PR #444, opened 2025-01-20
fix the default values for helm chart

runzhliu:


nccurry:
@runzhliu Thanks for the PR!
I updated your fork to the latest and updated the change to set the default explicitly in the values file, rather than rely on the default in the helm chart. 

This way, when we set `priorityClassName` to an empty string, we'll actually unset it in the rendered helm chart, rather than set it implicitly to `system-node-critical`

Issue #488, opened 2025-04-29
After MIG raise error: unsupported character in float while parsing


tongyangyeyue:
### What is the version?

4.1.1-4.0.4

### What happened?

when I enabling MIG on the gpu host , I noticed that some metrics are converted incorrectly and are incompatible with Prometheus. Examples include:
- DCGM_FI_DEV_CUDA_VISIBLE_DEVICES_STR
- DCGM_FI_DEV_MIG_MODE
- DCGM_FI_DEV_SM_CLOCK
The error message is 
```
"Error scraping target: unsupported character in float while parsing: "DCGM_FI_DEV_CUDA_VISIBLE_DEVICES_STR{gpu=\"0\",UUID=\"GPU-66068b5a-84f7-e3c9-bc10-e21eae988090\",pci_bus_id=\"00000000:09:00.0\",device=\"nvidia0\",modelName=\"NVIDIA H20\",GPU_I_PROFILE=\"2g.24gb\",GPU_I_ID=\"3\",Hostname=\"10-60-57-149\",DCGM_FI_DEV_MIG_CI_INFO=\"ERROR - FAILED TO CONVERT TO STRING\",DCGM_FI_DEV_MIG_GI_INFO=\"ERROR - FAILED TO CONVERT TO STRING\"} MIG-GPU-66068b5a-84f7-e3c9-bc10-e21eae988090/3"
```

![Image](https://github.com/user-attachments/assets/867a201f-637b-4a26-8b50-d25a573161a4)

### What did you expect to happen?

I will send a pull request to solve this problem， but I don't know how to convert this variable, I try to value.String() but result is empty

### What is the GPU model?

![Image](https://github.com/user-attachments/assets/09ac5fd7-1982-4f17-a913-7745cce07dc9)

![Image](https://github.com/user-attachments/assets/61182b59-2f07-48f8-a898-6c19c543be5e)

### What is the environment?

kubernetes

tongyangyeyue:
the pr is: https://github.com/NVIDIA/dcgm-exporter/pull/489

nccurry:
This is fixed in the latest version `4.6.0-4.8.3`

PR #489, opened 2025-04-29
fix(metrics): unsupported character in float while parsing

tongyangyeyue:
solve issue:  https://github.com/NVIDIA/dcgm-exporter/issues/488

when I enabling MIG on the gpu host , I noticed that some metrics are converted incorrectly and are incompatible with Prometheus. Examples include:

DCGM_FI_DEV_CUDA_VISIBLE_DEVICES_STR
DCGM_FI_DEV_MIG_MODE
DCGM_FI_DEV_SM_CLOCK


nccurry:
This should be fixed in the latest version `4.6.0-4.8.3`.

PR #505, opened 2025-06-03
Fix duplicate key in Helm chart

reyvonger:
Hello,  
when installing from the main branch, I encountered the following error:  

```
  Warning  InstallFailed     8s (x5 over 20s)  helm-controller  Helm install failed for release kube-system/nvdp-metrics with chart dcgm-exporter@4.1.2: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:  
  line 36: mapping key "app.kubernetes.io/component" already defined at line 35
```

This commit removes the duplicate key.


glowkey:
Can you outline the steps to reproduce the failure you're seeing?

PR #515, opened 2025-06-24
Kubernetes Pod Label Support

### LINKED ISSUE #423: Support collecting pod labels

New Feature

We only can collect DGCM metrics containing pod name / container name / namespace name, it is not possible to collect any labels.

### Feature Description

Hello,

We dig into dgcm exporter code base but there are no capability to collect pod labels and add these into metrics.

Do you have any plans supporting this ?

Best, 

Just adding an option to collect the labels from the pods and perhaps also specify which label like in kube state metrics https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-state-metrics/values.yaml#L342

### Additional context

_No response_

jlouazel:
## Summary

Adds opt-in Kubernetes pod label support to dcgm-exporter, enabling GPU metrics to be enriched with labels from pods using the resources. This allows tracking GPU usage by application, implementing chargeback, and debugging performance issues at the workload level.

We have been successfully running these changes in production since the end of December 2024, and it's long overdue that we share this back with the community.

## Implementation

Extends the existing `PodMapper` to fetch pod labels via the Kubernetes API and enrich GPU metrics with sanitized pod labels. Includes caching to minimize API server load.

### Key Features

* **Opt-in:** Disabled by default
* **Cached:** Prevents redundant API calls
* **RBAC integrated:** Conditional `ClusterRole`/`ClusterRoleBinding`
* **Backward compatible:** Existing metrics unchanged

**Configuration**

```yaml
kubernetes:
  enablePodLabels: true
  rbac:
    create: true
```

### Future Optimizations

The kubelet API could give faster, more scalable access to pod info, but it’s not part of the stable public interface. Currently, we rely on the Kubernetes API server, which incurs some overhead but benefits from caching. In the future, we may consider using the kubelet directly, although this approach comes with trade-offs regarding security and long-term support.

Fixes #423

### DIFF
--- a/deployment/templates/clusterrole.yaml
+++ b/deployment/templates/clusterrole.yaml
@@ -1,0 +1,12 @@
+{{- if and .Values.kubernetes.enablePodLabels .Values.kubernetes.rbac.create }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "dcgm-exporter.fullname" . }}-read-pods
+  labels:
+    {{- include "dcgm-exporter.labels" . | nindent 4 }}
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list"]
+{{- end }}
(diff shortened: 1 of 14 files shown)


jlouazel:
@rohit-arora-dev Do you have an ETA for the next release? We'd like to switch from our fork to the official version.

PR #538, opened 2025-08-12
(helm chart) Make priorityClassName optional

jyoungs:
The chart currently always sets priorityClassName on the DaemonSet by using an inline default-- and not actually accepting the value as input (which seems like a problem even if you do want to require it).

Keeping the same default, I'm allowing priorityClassName to be omitted, which simplifies the process of deploying this in a non-standard namespace.

This is basically a copy of #438 , but I squashed and signed (per CONTRIBUTING.md).


nccurry:
This is no longer needed in `4.6.0-4.8.3`

PR #535, opened 2025-08-07
DCGM-Exporter 4.3.1-4.4.0

glowkey:
- Update to DCGM 4.3.1
- Update to Go 1.24
- Enable DCGM_EXP_P2P_STATUS for reporting GPU peer-to-peer nvlink status
- Enable DRA support (@guptaNswati)
- Fix for empty HPC directory (@zhyocean)
- Enable Init container support (@hughobrien)

PR #531, opened 2025-07-30
Kubernetes Pod UID Support

andrew-leung:
# Kubernetes Pod UID Support

  ## Add Kubernetes pod UID support

  This commit adds an opt-in Kubernetes pod UID support to dcgm-exporter, enabling GPU metrics to be enriched with pod UID attributes from the pods that are using the resources. The implementation leverages the existing unified metadata caching system to minimize API server load and optimize performance.

  **Key Changes:**

  * Extended existing PodMapper with UID collection and caching capabilities
  * Integrated pod UIDs into GPU metrics as `pod_uid` attribute
  * Added `DCGM_EXPORTER_KUBERNETES_ENABLE_POD_UID` environment variable
  * Unified metadata caching for both labels and UIDs with optimized API calls
  * Comprehensive test coverage for UID functionality

  The pod UID is automatically included in GPU metrics when the feature is enabled, providing better observability and correlation between workloads and resource usage. This complements the existing pod labels feature.

  ## Add RBAC for Kubernetes pod UID integration

  This change extends the existing RBAC support to include pod UID collection alongside the pod labels feature. The implementation reuses the same `ClusterRole` and `ClusterRoleBinding` resources when either Kubernetes pod labels or pod UID features are enabled, ensuring RBAC resources are only created when metadata collection features are actually needed.

  The configuration includes `kubernetes.enablePodUID` to control the UID feature, building on the existing `kubernetes.rbac.create` setting for automatic RBAC resource management. When enabled, the daemonset receives the `DCGM_EXPORTER_KUBERNETES_ENABLE_POD_UID` environment variable to activate pod UID collection at runtime.

  **RBAC Changes:**
  * Updated conditional logic in ClusterRole and ClusterRoleBinding templates
  * Extended DaemonSet environment variables for UID feature
  * Added `kubernetes.enablePodUID` configuration option to values.yaml
  * Updated documentation to reflect support for both labels and UID features

  The changes maintain backwards compatibility and follow the same conditional patterns established by the pod labels feature, ensuring consistent behavior across both metadata collection capabilities.

Signed-off-by: Andrew Leung <andrew.leung@uber.com>

PR #543, opened 2025-08-15
Fix PodMetadata type issue resulting from merge conflict

andrew-leung:
# Fix: Resolve PodMetadata type mismatch from merge conflict

  ## Summary

  This PR fixes a compilation error introduced by a merge conflict between PR #531 (Kubernetes Pod UID Support) and PR #535 (DCGM-Exporter 4.3.1-4.4.0). The error manifests as a type mismatch where `labelCache` variable in the `toDeviceToPodsDRA` function has type `map[string]map[string]string` but function `createPodInfo` expects `map[string]PodMetadata`.

  ## Root Cause

  The issue was caused by concurrent development on the same code area. PR #535 was developed before PR #531's changes and used the old `labelCache` type (`map[string]map[string]string`) while calling the new `createPodInfo` function signature.

  ## Changes Made

  - **File**: `internal/pkg/transformation/kubernetes.go:378`
  - **Change**: Updated `labelCache` type in `toDeviceToPodsDRA` function from `map[string]map[string]string` to `map[string]PodMetadata`
  - **Reason**: Ensures type consistency with the `createPodInfo` function signature that was updated in PR #531

  ## Related PRs

  - **Introduced metadata changes**: PR #531 - Kubernetes Pod UID Support
  - **Introduced DRA function**: PR #535 - DCGM-Exporter 4.3.1-4.4.0 (fix dra labels when gpu is shared)
  - **Resolves**: Type mismatch between concurrent development branches

### DIFF
--- a/internal/pkg/transformation/kubernetes.go
+++ b/internal/pkg/transformation/kubernetes.go
@@ -375,7 +375,7 @@ func getSharedGPU(deviceID string) (string, bool) {
 
 func (p *PodMapper) toDeviceToPodsDRA(devicePods *podresourcesapi.ListPodResourcesResponse) map[string][]PodInfo {
 	deviceToPodsMap := make(map[string][]PodInfo)
-	labelCache := make(map[string]map[string]string) // Cache to avoid duplicate API calls
+	labelCache := make(map[string]PodMetadata) // Cache to avoid duplicate API calls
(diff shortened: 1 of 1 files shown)


glowkey (approved):
thanks!

Issue #530, opened 2025-07-23
Not collecting GPU metrics; error getting devices count: Cannot perform the requested operation because NVML doesn't exist on this system.


pomyslowynick:
### What is the version?

4.2.3-4.2.0

### What happened?

After upgrading to the newest version from `3.3.9-3.6.1` to `4.2.3-4.2.0` we are seeing an "Unknown error" on initializing the `NVML` library.
I can see from the debugger that this error is returned by this line [in nvml-go](https://github.com/NVIDIA/go-nvml/blob/main/pkg/nvml/nvml.go?plain=1#L33), but couldn't see much more since then it just call the C library.

We are runing the exporter in a GKE cluster with a default GPU driver installed on the node.
Attaching the full log:

```
nvidia-dcgm-exporter-s9vsg time=2025-07-23T11:26:35.584Z level=DEBUG msg="Debug output is enabled"
nvidia-dcgm-exporter-s9vsg time=2025-07-23T11:26:35.585Z level=INFO msg="Starting dcgm-exporter" Version=4.2.3-4.2.0
nvidia-dcgm-exporter-s9vsg time=2025-07-23T11:26:35.680Z level=INFO msg="Attempting to connect to remote hostengine at 10.253.44.117"
nvidia-dcgm-exporter-s9vsg time=2025-07-23T11:26:35.686Z level=INFO msg="Initialized DCGM Fields module."
nvidia-dcgm-exporter-s9vsg time=2025-07-23T11:26:35.686Z level=INFO msg="Attempting to initialize NVML library."
nvidia-dcgm-exporter-s9vsg time=2025-07-23T11:26:35.781Z level=ERROR msg="Cannot init NVML library; err: Unknown Error"
nvidia-dcgm-exporter-s9vsg time=2025-07-23T11:26:35.781Z level=INFO msg="DCGM successfully initialized!"
nvidia-dcgm-exporter-s9vsg time=2025-07-23T11:26:35.781Z level=INFO msg="NVML provider successfully initialized!"
nvidia-dcgm-exporter-s9vsg time=2025-07-23T11:26:38.088Z level=INFO msg="Not collecting DCP metrics: This request is serviced by a module of DCGM that is not currently loaded"
nvidia-dcgm-exporter-s9vsg time=2025-07-23T11:26:38.088Z level=INFO msg="Falling back to metric file '/etc/dcgm-exporter/metrics.csv'"
[...]

PR #557, opened 2025-09-10
feat(dcgm-exporter): add option to fail on nvml provider init error

daveoy:
this PR adds the option to exit if we are unable to initialize the nvml provider. it has been observed and reported that nvml.Init() periodically fails to initialize the provider, but this error is currently ignored in dcgm-exporter. the result is an exporter that is unable to export gpu metrics, and in my case, exports nothing. the only indication that there has been an issue on startup is visible in the logs.

nvml init issue reported here https://github.com/NVIDIA/go-nvml/issues/164 and here https://github.com/NVIDIA/dcgm-exporter/issues/530

### DIFF
--- a/pkg/cmd/app.go
+++ b/pkg/cmd/app.go
@@ -82,7 +82,6 @@ const (
 	CLIConfigMapData              = "configmap-data"
 	CLIWebSystemdSocket           = "web-systemd-socket"
 	CLIWebConfigFile              = "web-config-file"
-	CLIFailOnNVMLInitError        = "fail-on-nvml-init-error"
 	CLIXIDCountWindowSize         = "xid-count-window-size"
 	CLIReplaceBlanksInModelName   = "replace-blanks-in-model-name"
 	CLIDebugMode                  = "debug"
@@ -224,12 +223,6 @@ func NewApp(buildVersion ...string) *cli.App {
 			Value:   "",
 			Usage:   "Web configuration file following webConfig spec: https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md.",
 			EnvVars: []string{"DCGM_EXPORTER_WEB_CONFIG_FILE"},
-		},
-		&cli.BoolFlag{
-			Name:    CLIFailOnNVMLInitError,
-			Value:   false,
-			Usage:   "Fail the application if NVML initialization fails",
-			EnvVars: []string{"DCGM_EXPORTER_FAIL_ON_NVML_INIT_ERROR"},
 		},
 		&cli.IntFlag{
 			Name:    CLIXIDCountWindowSize,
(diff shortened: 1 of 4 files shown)


daveoy:
we have confirmed internally that this feature works.

by enabling the feature using the env var, and removing the security context the pod successfully reaches crashloop backoff, rather than sitting there running but exporting zero metrics.

logs:
```
% kubectl logs zz-test-dcgm
Warning #2: dcgm-exporter doesn't have sufficient privileges to expose profiling metrics. To get profiling metrics with dcgm-exporter, use --cap-add SYS_ADMIN
time=2025-09-10T16:44:04.147Z level=INFO msg="Starting dcgm-exporter" Version=4.4.0-4.5.0
time=2025-09-10T16:44:04.149Z level=INFO msg="Attempting to initialize DCGM."
time=2025-09-10T16:44:04.152Z level=INFO msg="Initialized DCGM Fields module."
time=2025-09-10T16:44:04.152Z level=INFO msg="Attempting to initialize NVML library."
time=2025-09-10T16:44:04.152Z level=ERROR msg="Cannot init NVML library; err: ERROR_LIBRARY_NOT_FOUND"
time=2025-09-10T16:44:04.152Z level=ERROR msg=ERROR_LIBRARY_NOT_FOUND
```

daveoy:
cc @glowkey (sorry friend, you're the only contrib i know the github handle of)

glowkey:
This looks good. I think it would be useful to enable this by default rather than only with a flag. Conversely, the behavior could then be disabled based on the flag introduced in this MR: https://github.com/NVIDIA/dcgm-exporter/pull/555. If you agree, could you refactor in this way?

daveoy:
Yep, happy to make that change. I’ll push it today.

pkg/cmd/app.go
```diff
@@ -422,6 +429,7 @@ func startDCGMExporter(c *cli.Context) error {
 
 		// Only validate prerequisites if not disabled.
 		if !config.DisableStartupValidate {
+			config.FailOnNVMLInitError = true // enable failure on NVML Init Error
```
daveoy, line 432: @glowkey this means that if DisableStartupValidate is false, meaning we want startup validating, we will exit on NVML init error

if DisableStartupValidate is true, meaning we want to start up without validating rules (and now, nvml init errors), we can proceed with FailOnNVMLInitError = false (default).

this means users can also decide to enable DisableStartupValidate, but also enable FailOnNVMLInitError, so no rules are validated via prerequisites, but the exporter will bail if nvml init fails.

daveoy: would we rather this be enforced by DisableStartupValidate? the only disadvantage there is that people cant decide they want to skip validating the rules AND also bail on nvml init error.

glowkey (approved):

PR #558, opened 2025-09-16
Update README.md golang requirement

Pennswood:
go.mod requires golang 1.24 but documentation still says 1.22


Pennswood:
This might be seperate: why do we .gitignore `dcgm-exporter` as this is might be causing cmd\dcgm-exporter\main.go to become ignored I think.

glowkey (approved):

PR #564, opened 2025-10-04
Add allow list for pod label filtering

andrew-leung:
## Summary

  Adds support for filtering Kubernetes pod labels using regex patterns to prevent high-cardinality labels from causing metric explosion. This feature can help avoid copying high cardinality or unnecessarily large numbers of pod labels into metrics.

  ## Changes

  - Added `--kubernetes-pod-label-allowlist-regex` CLI flag and `DCGM_EXPORTER_KUBERNETES_POD_LABEL_ALLOWLIST_REGEX` environment variable
  - Implemented two-tier caching (pre-compiled regexes + lookup cache) to minimize performance overhead
  - Updated Helm chart with `kubernetes.podLabelAllowlistRegex` configuration option
  - Empty allowlist maintains backward compatibility (includes all labels)

  ## Configuration Examples

  ```yaml
  # Helm values.yaml
  kubernetes:
    enablePodLabels: true
    podLabelAllowlistRegex:
      - "^app$"
      - "^app\\.kubernetes\\.io/.*"
      - "^(tier|environment|version)$"

  # Environment variable
  DCGM_EXPORTER_KUBERNETES_POD_LABEL_ALLOWLIST_REGEX="^app$,^tier$"

  # CLI flag
  dcgm-exporter --kubernetes-pod-label-allowlist-regex "^app$" --kubernetes-pod-label-allowlist-regex "^tier$"
  ```

   ## Testing

  Added unit tests covering cache initialization, filtering logic, and integration with pod metadata retrieval.
  Tested build on h100 internal host.


internal/pkg/transformation/kubernetes_label_filter_test.go
```diff
@@ -0,0 +1,427 @@
...
+			// Call getPodMetadata
+			metadata, err := podMapper.getPodMetadata("default", "test-pod")
+			require.NoError(t, err, "getPodMetadata should not return error")
+			require.NotNil(t, metadata, "metadata should not be nil")
+
+			// Verify UID is set correctly
+			assert.Equal(t, "test-uid-123", metadata.UID, "UID should match")
+
+			// Verify labels match expected (after filtering and sanitization)
+			assert.Equal(t, len(tt.expectedLabels), len(metadata.Labels),
```
MKrupauskas, line 407:
```suggestion
			assert.Equal(t, tt.expectedLabels, metadata.Labels)
```

Do we need all of these assertion loops if we can just assert that the map we expect is the map we get? That should make for a more robust assertion

amitaekbote (approved):
🚢

internal/pkg/transformation/kubernetes_label_filter_test.go
andrew-leung: Good call, cleaned up these assertions.

glowkey (approved):

PR #574, opened 2025-10-23
Fix unbounded memory growth in pod label allowlist cache

andrew-leung:
### Problem
The pod label allowlist cache used an unbounded `sync.Map` that would grow indefinitely in clusters with high pod churn or diverse label patterns. Over days/weeks, this could consume hundreds of MBs or GBs of memory.

### Solution
Replaced with a bounded LRU cache (using `container/list` + map) that:
- Evicts least recently used entries when full
- Maintains O(1) lookups and updates
- Thread-safe with mutex protection

### Configuration
Added `--kubernetes-pod-label-cache-size` flag (env: `DCGM_EXPORTER_KUBERNETES_POD_LABEL_CACHE_SIZE`)
- Default: 150,000 entries (~18MB)
- Adjustable based on cluster size and memory constraints

### Testing
- Updated existing label filter tests
- Added comprehensive LRU-specific tests covering:
  - Eviction behavior
  - LRU ordering correctness
  - Concurrent access safety

### DIFF
--- a/internal/pkg/appconfig/types.go
+++ b/internal/pkg/appconfig/types.go
@@ -45,6 +45,7 @@ type Config struct {
 	KubernetesEnablePodUID           bool
 	KubernetesGPUIdType              KubernetesGPUIDType
 	KubernetesPodLabelAllowlistRegex []string // Regex patterns for filtering pod labels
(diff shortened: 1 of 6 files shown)


glowkey (approved):
LGTM. Thanks!

PR #584, opened 2025-11-18
DCGM-Exporter release 4.4.2-4.7.0

glowkey:
- Security improvements
- Crash if collector cannot be initialized under startup validation
- Fix unbounded label cache size
- Track unallocated GPUs


rohit-arora-dev (approved):
/LGTM

Issue #549, opened 2025-08-29
GPU Sharing metrics empty with KUBERNETES_VIRTUAL_GPUS


dcendents:
### What is the version?

4.2.3-4.1.3

### What happened?

When I configure dcgm-exporter with `KUBERNETES_VIRTUAL_GPUS` set to `true` and no pods are using the GPU on a node with Time-Slicing, no metrics are returned at all.

### What did you expect to happen?

I would expect to get usage metrics of 0, get the GPU temperature, power consumptions, etc.

### What is the GPU model?

Quadro-K2200 (testing using an old workstation)

### What is the environment?

Kubernetes cluster 1.32
The GPU node is Ubuntu 24.04

### How did you deploy the dcgm-exporter and what is the configuration?

Through gpu-operator helm chart, everything is default values except added env var `KUBERNETES_VIRTUAL_GPUS`

### How to reproduce the issue?

Do not execute any pod on the node except for the system daemon sets.
e.g.:
```
NAMESPACE        NAME                                                      READY   STATUS      RESTARTS       AGE
cilium-test-1    host-netns-wmk8k                                          1/1     Running     0              25h
gpu-management   gpu-feature-discovery-ps2gz                               2/2     Running     0              20m
gpu-management   nvidia-cuda-validator-x9d8s                               0/1     Completed   0              20m
gpu-management   nvidia-dcgm-exporter-lb6zk                                1/1     Running     0              20m
gpu-management   nvidia-device-plugin-daemonset-kqf65                      2/2     Running     0              20m
gpu-management   nvidia-gpu-operator-node-feature-discovery-worker-t6ldt   1/1     Running     2 (3d2h ago)   10d
gpu-management   nvidia-operator-validator-c2rkw                           1/1     Running     0              20m
kube-system      cilium-envoy-5gzvr                                        1/1     Running     0              3d2h
kube-system      cilium-pwjtg                                              1/1     Running     0              3d2h
kube-system      csi-smb-node-g5pj7                                        3/3     Running     0              3d2h
monitoring       kube-prometheus-stack-prometheus-node-exporter-ft2tk      1/1     Running     0              3d2h
```

### Anything else we need to know?

If I port-forward the dcgm-exporter port locally here is what I get:

```
curl -sL http://127.0.0.1:9401/metrics
# HELP DCGM_FI_DEV_SM_CLOCK SM clock frequency (in MHz).
# TYPE DCGM_FI_DEV_SM_CLOCK gauge
[...]

PR #594, opened 2025-12-04
Add per-process GPU metrics for time-sharing and MIG

krystiancastai:
### Problem

When using GPU time-sharing (multiple pods sharing a single GPU or MIG instance), DCGM Exporter metrics were inaccurate:
  
**Regular GPU time-sharing:** 
Device-level metrics were duplicated to each pod, making it impossible to determine actual per-pod resource consumption:
  ```
  DCGM_FI_DEV_GPU_UTIL{pod="gpu-workload-1",...} 99
  DCGM_FI_DEV_GPU_UTIL{pod="gpu-workload-2",...} 99
  DCGM_FI_DEV_GPU_UTIL{pod="gpu-workload-3",...} 99
  DCGM_FI_DEV_GPU_UTIL{pod="gpu-workload-4",...} 99
  ```

**MIG + time-sharing:** 
Only one random pod was reported per MIG instance - other pods sharing the same MIG instance were missing from metrics entirely.

  ### Solution

  This PR adds **per-process GPU metrics collection** using NVML APIs, enabling per-pod metrics in time-sharing scenarios:

  - Uses `nvmlDeviceGetComputeRunningProcesses` to get per-process memory usage
  - Uses `nvmlDeviceGetProcessUtilization` to get per-process SM utilization (regular GPUs only)
  - Maps process PIDs to Kubernetes pods via cgroup
  - Supports both regular GPUs and MIG instances

  ### Configuration

The changes in this PR only affect the code path when the `--kubernetes-virtual-gpus` flag is enabled (or `KUBERNETES_VIRTUAL_GPUS=true` environment variable). When this flag is disabled, the existing behavior remains unchanged.

  Required environment variables:

  - `KUBERNETES_VIRTUAL_GPUS=true`
  - `DCGM_EXPORTER_KUBERNETES=true`
  - `DCGM_EXPORTER_KUBERNETES_ENABLE_POD_UID=true`
  - `DCGM_EXPORTER_KUBERNETES_GPU_ID_TYPE` — must match your cloud provider:
    - GKE (GCP): `device-name`
    - EKS (AWS): `uid`

  Required pod spec setting:
  - `hostPID=true`
  - `securityContext.privileged=true`

  ---

  ### Per-Process Metrics by GPU Type

  #### Regular GPUs (Time-Sharing)

  Both GPU utilization and memory metrics are available per-process:

  | Metric | Description |
  |--------|-------------|
  | `DCGM_FI_DEV_GPU_UTIL` | Per-process SM utilization (%) |
[...]

PR #644, opened 2026-03-13
fix: account for partially-allocated gpus

pintohutch:
Fixes the issue mentioned in
https://github.com/NVIDIA/dcgm-exporter/issues/549#issuecomment-3736684867.

Given KUBERNETES_VIRTUAL_GPUS is enabled:
When the Kubernetes pod-mapping builds the set of new metrics, we previously ignored the case where some devices were allocated, while others were not. So we add a check to each device lookup, where if there are no associated pods, we copy over the "pod-less" metrics to the final map.

### DIFF
--- a/internal/pkg/transformation/kubernetes.go
+++ b/internal/pkg/transformation/kubernetes.go
@@ -269,33 +269,37 @@ func (p *PodMapper) Process(metrics collector.MetricsByCounter, deviceInfo devic
 				}
 
 				podInfos := deviceToPods[deviceID]
-				for _, pi := range podInfos {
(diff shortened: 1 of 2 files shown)


rvatkar:
Can you rebase this on the latest in main?

rvatkar:
@pintohutch thanks for this contribution.
After reviewing, we've found that [PR #594](https://github.com/NVIDIA/dcgm-exporter/pull/594) (already merged to main) already addresses this use case. We're planning to include it in the next release.

We're closing this PR since the functionality is covered

PR #649, opened 2026-04-01
Add vsock support for serving metrics to hypervisors

machadovilaca:
Add an optional AF_VSOCK listener that serves the same metrics endpoint over virtio-vsock, enabling guest to host metrics collection without network dependencies.

In GPU passthrough and vGPU use cases on KubeVirt (and other hypervisors), the guest VM is often not configured with a network path to the host, but needs GPU telemetry from dcgm-exporter running inside the guest.

### DIFF
--- a/pkg/cmd/app.go
+++ b/pkg/cmd/app.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"math"
 	"os"
 	"os/signal"
 	"runtime"
@@ -102,6 +103,7 @@ const (
 	CLIDisableStartupValidate           = "disable-startup-validate"
 	CLIEnableGPUBindUnbindWatch         = "enable-gpu-bind-unbind-watch"
 	CLIGPUBindUnbindPollInterval        = "gpu-bind-unbind-poll-interval"
+	CLIVSOCKPort                        = "vsock-port"
 )
 
 func NewApp(buildVersion ...string) *cli.App {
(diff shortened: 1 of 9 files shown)


nccurry:
Can you rebases this on the latest in main

pkg/cmd/app.go
```diff
@@ -1075,6 +1083,11 @@ func contextToConfig(c *cli.Context) (*appconfig.Config, error) {
 		return nil, fmt.Errorf("invalid %s parameter value: %s", CLIDCGMLogLevel, dcgmLogLevel)
 	}
 
+	if c.Uint(CLIVSOCKPort) > math.MaxUint32 {
+		return nil, fmt.Errorf("vsock port must be in range 1-65535")
```
nccurry, line 1087: Is this the right error message for Uint32?
Maybe something like `return nil, fmt.Errorf("vsock port must fit in 32 bits")`

nccurry:
Can you look in to using a socat sidecar, or one of the other methods for proxying to vock, for this? 
e.g. 
```yaml
  containers:
    - name: dcgm-exporter
      image: nvcr.io/nvidia/k8s/dcgm-exporter:...
      ports:
        - containerPort: 9400
    - name: vsock-proxy
      image: alpine/socat
      command: ["socat", "VSOCK-LISTEN:9400,reuseaddr,fork", "TCP:localhost:9400"]
```

DCGM Exporter is really intended to channel DCGM input into Prometheus output. 

Adding an entire VSOCK endpoint is not an insignificant change to the API for something that could be done externally.

PR #655, opened 2026-04-29
Rename Hostname label to hostname

daphne2337:
All the other 500+ labels are all uppercase or lowercase
Joining it with slurm-derived metrics (label hostname identical except for case) activates weird Grafana bugs such as duplicate timeseries

Tested: DISABLE_STARTUP_VALIDATE=true go test ./internal/pkg/rendermetrics/...


nccurry (approved):
LGTM

PR #663, opened 2026-05-29
collector: allocate a fresh label map per entity in experimental collectors

rocker-zhang:
## Problem

The experimental collectors (XID, clock events, GPU health, P2P status)
build a single label map before the per-entity loop and reuse it for every
GPU. `getLabelsFromCounters` only adds label keys, so keys set for one GPU
leak onto GPUs that lack them.

In `expcollector.go` the idle/zero-value branch passes that shared map
directly to `createMetric`, which stores it by reference (`Labels: labels`).
The next entity's `getLabelsFromCounters` then mutates the same map, so an
idle GPU ends up reporting another GPU's label values.

## Fix

Allocate the label map inside the per-entity loop in `expcollector.go`,
`gpu_health_collector.go`, and `p2p_status_collector.go`, so each entity
starts with its own labels.

## Test

`Test_xidCollector_GetMetrics_PerEntityLabelIsolation` sets up two GPUs with
distinct label values where only one GPU has an XID event, forcing the other
down the idle branch. It fails on `main` (the idle GPU reports the other
GPU's label) and passes with this change.

```
go test ./internal/pkg/collector/
ok  github.com/NVIDIA/dcgm-exporter/internal/pkg/collector
```

### DIFF
--- a/internal/pkg/collector/expcollector.go
+++ b/internal/pkg/collector/expcollector.go
@@ -70,9 +70,6 @@ func (c *expCollector) getMetrics() (MetricsByCounter, error) {
 		}
 	}
 
-	labels := map[string]string{}
-	labels[windowSizeInMSLabel] = fmt.Sprint(c.windowSize)
-
 	monitoringInfo := devicemonitoring.GetMonitoredEntities(c.deviceWatchList.DeviceInfo())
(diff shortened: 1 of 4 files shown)


nccurry:

PR #664, opened 2026-05-29
Add v1 api support for dra-driver-nvidia-gpu

### LINKED ISSUE #590: Add support for K8s v1.34 resource.k8s.io/v1 DRA APIs

guptaNswati:
This PR adds proper API discovery, refactors the DRA cache to use a Lister, and fixes MIG correctness gaps in pod-to-metric attribution for DRA.

Fixes https://github.com/NVIDIA/dcgm-exporter/issues/590. 

Tested both full GPU and MIG paths. Sharing test logs on A30 with MIG enabled. 

```
time=2026-06-02T19:42:06.979Z level=INFO msg="Initializing Pod Informer" nodeName=nim-operator-72q5r53
time=2026-06-02T19:42:07.068Z level=INFO msg="Using ResourceSlice API" groupVersion=resource.k8s.io/v1
time=2026-06-02T19:42:07.168Z level=INFO msg="Started DRAResourceSliceManager"

DCGM_FI_DEV_SM_CLOCK{gpu="0",UUID="GPU-f50641a9-38d7-6df1-0bec-e661b4ca1847",pci_bus_id="00000000:3B:00.0",device="nvidia0",modelName="NVIDIA A30",hostname="nim-operator-72q5r53"} 210

DCGM_FI_DEV_SM_CLOCK{gpu="1",UUID="GPU-e04ce2a0-21a4-b68c-854a-f5c0d26bfc0b",pci_bus_id="00000000:D8:00.0",device="nvidia1",modelName="NVIDIA A30",GPU_I_PROFILE="1g.6gb",GPU_I_ID="3",hostname="nim-operator-72q5r53",container="ctr0",dra_claim_name="pod-c98945f5d-dnjs6-mig-devices-thfdj",dra_claim_namespace="gpu-test4",dra_device_name="gpu-1-mig-1g6gb-14-0",dra_driver_name="gpu.nvidia.com",dra_mig_device_uuid="MIG-c73ab1c4-6cb7-59b6-be7e-21c4ac35c3b6",dra_mig_profile="1g.6gb",dra_pool_name="nim-operator-72q5r53",namespace="gpu-test4",pod="pod-c98945f5d-dnjs6"} 1440
DCGM_FI_DEV_SM_CLOCK{gpu="1",UUID="GPU-e04ce2a0-21a4-b68c-854a-f5c0d26bfc0b",pci_bus_id="00000000:D8:00.0",device="nvidia1",modelName="NVIDIA A30",GPU_I_PROFILE="1g.6gb",GPU_I_ID="3",hostname="nim-operator-72q5r53",container="ctr1",dra_claim_name="pod-c98945f5d-dnjs6-mig-devices-thfdj",dra_claim_namespace="gpu-test4",dra_device_name="gpu-1-mig-1g6gb-14-0",dra_driver_name="gpu.nvidia.com",dra_mig_device_uuid="MIG-c73ab1c4-6cb7-59b6-be7e-21c4ac35c3b6",dra_mig_profile="1g.6gb",dra_pool_name="nim-operator-72q5r53",namespace="gpu-test4",pod="pod-c98945f5d-dnjs6"} 1440
DCGM_FI_DEV_SM_CLOCK{gpu="1",UUID="GPU-e04ce2a0-21a4-b68c-854a-f5c0d26bfc0b",pci_bus_id="00000000:D8:00.0",device="nvidia1",modelName="NVIDIA A30",GPU_I_PROFILE="1g.6gb",GPU_I_ID="3",hostname="nim-operator-72q5r53",container="ctr2",dra_claim_name="pod-c98945f5d-dnjs6-mig-devices-thfdj",dra_claim_namespace="gpu-test4",dra_device_name="gpu-1-mig-1g6gb-14-0",dra_driver_name="gpu.nvidia.com",dra_mig_device_uuid="MIG-c73ab1c4-6cb7-59b6-be7e-21c4ac35c3b6",dra_mig_profile="1g.6gb",dra_pool_name="nim-operator-72q5r53",namespace="gpu-test4",pod="pod-c98945f5d-dnjs6"} 1440
DCGM_FI_DEV_SM_CLOCK{gpu="1",UUID="GPU-e04ce2a0-21a4-b68c-854a-f5c0d26bfc0b",pci_bus_id="00000000:D8:00.0",device="nvidia1",modelName="NVIDIA A30",GPU_I_PROFILE="1g.6gb",GPU_I_ID="3",hostname="nim-operator-72q5r53",container="ctr3",dra_claim_name="pod-c98945f5d-dnjs6-mig-devices-thfdj",dra_claim_namespace="gpu-test4",dra_device_name="gpu-1-mig-1g6gb-14-0",dra_driver_name="gpu.nvidia.com",dra_mig_device_uuid="MIG-c73ab1c4-6cb7-59b6-be7e-21c4ac35c3b6",dra_mig_profile="1g.6gb",dra_pool_name="nim-operator-72q5r53",namespace="gpu-test4",pod="pod-c98945f5d-dnjs6"} 1440
```

### DIFF
--- a/internal/pkg/transformation/dra.go
+++ b/internal/pkg/transformation/dra.go
@@ -22,8 +22,13 @@ import (
 	"log/slog"
 	"time"
 
+	resourcev1 "k8s.io/api/resource/v1"
 	resourcev1beta1 "k8s.io/api/resource/v1beta1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	resourcelistersv1 "k8s.io/client-go/listers/resource/v1"
+	resourcelistersv1beta1 "k8s.io/client-go/listers/resource/v1beta1"
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/NVIDIA/dcgm-exporter/internal/pkg/kubeclient"
@@ -31,7 +36,65 @@ import (
 
 const (
 	informerResyncPeriod = 10 * time.Minute
+
+	// resourceSliceCacheSyncTimeout bounds how long we wait for the initial
+	// ResourceSlice informer cache sync. Discovery may report that v1 (or
+	// v1beta1) is served while RBAC denies list/watch on resourceslices; in
+	// that case the informer's reflector retries forever and HasSynced never
+	// returns true. A bounded wait prevents the exporter from blocking on
+	// startup in such misconfigurations.
+	resourceSliceCacheSyncTimeout = 30 * time.Second
 )
+
+// resourceSliceAPIVersion identifies which ResourceSlice API version the cluster serves.
+type resourceSliceAPIVersion int
+
+const (
+	resourceSliceAPIUnknown resourceSliceAPIVersion = iota
+	resourceSliceAPIV1
+	resourceSliceAPIV1beta1
+)
+
+func (v resourceSliceAPIVersion) String() string {
+	switch v {
+	case resourceSliceAPIV1:
+		return resourcev1.SchemeGroupVersion.String()
+	case resourceSliceAPIV1beta1:
+		return resourcev1beta1.SchemeGroupVersion.String()
+	default:
+		return "unknown"
+	}
+}
+
+// detectResourceSliceAPIVersion returns the ResourceSlice API version served by
+// the cluster. It uses discovery to look for the resourceslices resource under
+// resource.k8s.io/v1 first (stable, Kubernetes 1.34+) and falls back to
+// resource.k8s.io/v1beta1 (older clusters). Returns resourceSliceAPIUnknown if
+// neither GroupVersion is served.
+func detectResourceSliceAPIVersion(client kubernetes.Interface) resourceSliceAPIVersion {
+	apiVersion := []struct {
+		groupVersion string
+		version      resourceSliceAPIVersion
+	}{
+		{resourcev1.SchemeGroupVersion.WithResource("resourceslices").GroupVersion().String(), resourceSliceAPIV1},
+		{resourcev1beta1.SchemeGroupVersion.WithResource("resourceslices").GroupVersion().String(), resourceSliceAPIV1beta1},
+	}
+
+	for _, a := range apiVersion {
+		resources, err := client.Discovery().ServerResourcesForGroupVersion(a.groupVersion)
+		if err != nil {
+			// Discovery returns an error when the group/version isn't served.
+			slog.Debug("ResourceSlice discovery failed", "groupVersion", a.groupVersion, "error", err)
+			continue
+		}
+		for _, r := range resources.APIResources {
+			if r.Name == "resourceslices" {
+				return a.version
+			}
+		}
+	}
+	return resourceSliceAPIUnknown
+}
 
 func NewDRAResourceSliceManager() (*DRAResourceSliceManager, error) {
 	client, err := kubeclient.GetKubeClient()
@@ -39,37 +102,47 @@ func NewDRAResourceSliceManager() (*DRAResourceSliceManager, error) {
 		return nil, fmt.Errorf("error getting kube client: %w", err)
 	}
 
+	apiVersion := detectResourceSliceAPIVersion(client)
+	if apiVersion == resourceSliceAPIUnknown {
+		return nil, fmt.Errorf("ResourceSlice API not served by cluster (looked for %s and %s)",
+			resourcev1.SchemeGroupVersion, resourcev1beta1.SchemeGroupVersion)
+	}
+
 	factory := informers.NewSharedInformerFactory(client, informerResyncPeriod)
-	informer := factory.Resource().V1beta1().ResourceSlices().Informer()
 
-	m := &DRAResourceSliceManager{
-		factory:      factory,
-		informer:     informer,
-		deviceToUUID: make(map[string]string),
-		migDevices:   make(map[string]*DRAMigDeviceInfo),
+	m := &DRAResourceSliceManager{factory: factory}
+
+	var hasSynced cache.InformerSynced
+	switch apiVersion {
+	case resourceSliceAPIV1:
+		informer := factory.Resource().V1().ResourceSlices().Informer()
+		lister := factory.Resource().V1().ResourceSlices().Lister()
+		m.lookup = makeV1Lookup(lister)
+		hasSynced = informer.HasSynced
+	case resourceSliceAPIV1beta1:
+		informer := factory.Resource().V1beta1().ResourceSlices().Informer()
+		lister := factory.Resource().V1beta1().ResourceSlices().Lister()
+		m.lookup = makeV1beta1Lookup(lister)
+		hasSynced = informer.HasSynced
 	}
 
-	_, err = informer.AddEventHandler(&cache.FilteringResourceEventHandler{
-		FilterFunc: func(obj interface{}) bool {
-			s := obj.(*resourcev1beta1.ResourceSlice)
-			return s.Spec.Driver == DRAGPUDriverName
-		},
-		Handler: cache.ResourceEventHandlerFuncs{
-			AddFunc:    m.onAddOrUpdate,
-			UpdateFunc: func(_, o interface{}) { m.onAddOrUpdate(o) },
-			DeleteFunc: m.onDelete,
-		},
-	})
-	if err != nil {
-		return nil, fmt.Errorf("error adding event handler: %w", err)
-	}
+	slog.Info("Using ResourceSlice API", "groupVersion", apiVersion.String())
 
 	ctx, cancel := context.WithCancel(context.Background())
 	m.cancelContext = cancel
 	factory.Start(ctx.Done())
 
-	if !cache.WaitForCacheSync(ctx.Done(), informer.HasSynced) {
+	// Bound the initial sync. Discovery may say the API is served, but RBAC
+	// can still deny list/watch on resourceslices, in which case HasSynced
+	// would never become true and startup would hang.
+	syncCtx, syncCancel := context.WithTimeout(ctx, resourceSliceCacheSyncTimeout)
+	defer syncCancel()
+	if !cache.WaitForCacheSync(syncCtx.Done(), hasSynced) {
 		cancel()
+		if syncCtx.Err() == context.DeadlineExceeded {
+			return nil, fmt.Errorf("ResourceSlice informer cache sync timed out after %s "+
+				"(check RBAC: list/watch on resource.k8s.io/resourceslices)", resourceSliceCacheSyncTimeout)
+		}
 		return nil, fmt.Errorf("ResourceSlice informer cache sync failed")
 	}
 	return m, nil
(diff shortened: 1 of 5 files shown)


guptaNswati:
I will add tests for api discovery and others but its ready for review. 

Edited to mark this done.

guptaNswati:
cc @varunrsekar @nccurry

internal/pkg/transformation/kubernetes_test.go
```diff
@@ -701,7 +698,10 @@ func TestPodDRAInfo(t *testing.T) {
 				}},
 			}
 
-			got := pm.toDeviceToPodsDRA(resp)
+			// deviceInfo is nil here: the existing cases don't exercise the
+			// MIG path. NVML lookup is skipped when deviceInfo is nil,
+			// so the function falls back to parent-UUID-only mapping
+			got := pm.toDeviceToPodsDRA(resp, nil)
```
nccurry, line 704: Could we add direct coverage for the DRA MIG GPU-instance mapping path?

The current DRA test intentionally calls `toDeviceToPodsDRA(resp, nil)`, so it only verifies the parent-UUID fallback and skips `draMIGGPUInstanceIdentifier`. That leaves the main MIG correctness path unproven: DRA MIG claims should be registered under both the parent GPU UUID and DCGM’s per-MIG metric key (`<gpuIndex>-<gpuInstanceID>`, e.g. `0-3`), so MIG metrics can receive pod/DRA labels.

A focused unit test could mock NVML to return `{ParentUUID: "GPU-parent-uuid", GPUInstanceID: 3}`, mock `deviceinfo.Provider` so that parent UUID maps to GPU index `0`, then assert that `toDeviceToPodsDRA` returns entries for both `"GPU-parent-uuid"` and `"0-3"` with the expected pod and `DynamicResources.MIGInfo`.

Something like `TestPodDRAInfo_MIGAddsGPUInstanceMapping` would close this gap.

internal/pkg/transformation/kubernetes.go
```diff
@@ -603,18 +603,21 @@ func (p *PodMapper) toDeviceToPodsDRA(devicePods *podresourcesapi.ListPodResourc
...
+						// Build the set of DRA mapping keys under which this
+						// pod/container must be discoverable. DCGM emits
+						// MIG-instance metrics keyed by "<gpuIndex>-<gpuInstanceID>"
+						// (see Metric.GetIDOfType), so for MIG claims we
+						// additionally register the GPU-instance identifier in
+						// addition to the parent UUID. Without this, per-MIG
+						// metrics never receive pod attributes.
+						draMappingKeys := []string{mappingKey}
+						if migInfo != nil {
+							if giID := p.draMIGGPUInstanceIdentifier(migInfo, deviceInfo); giID != "" {
```
nccurry, line 615: If we want to cover the full enrichment path, a second test could run `PodMapper.Process` with `KubernetesEnableDRA=true` and a MIG metric whose `GetIDOfType` resolves to `0-3`, then assert the output metric has `pod`, `namespace`, `container`, `dra_*`, and `dra_mig_*` attributes. That would prove the mapping is not only created but actually used during metric enrichment.

internal/pkg/transformation/kubernetes_test.go
guptaNswati: done

internal/pkg/transformation/kubernetes.go
guptaNswati: done

rvatkar:
Thanks for updating this! I checked #664 against our internal dcgm exporter work.

The MIG attribution goal is already covered on our mainline, but there is one behavior difference we should align on: our code maps a DRA MIG claim to the GI key only, and drops the mapping if NVML cannot resolve the GPU-instance ID. This PR registers both the parent UUID and the GI key, and falls back to parent-UUID-only on lookup failure. I think that
drop vs. degrade behavior is the main thing to reconcile.

The v1 DRA API support is still not addressed internally on main: we do not have the v1/v1beta1 discovery fallback or the bounded cache-sync timeout. Those pieces are definitely valuable. One caveat from our side: the informer-to-Lister refactor also changes lookup from our current O(1) map lookup to a per-claim scan of the full cached ResourceSlice list, so we would prefer landing the v1 discovery on the existing event-handler model if possible.

internal/pkg/transformation/dra.go
```diff
@@ -79,100 +152,123 @@ func (m *DRAResourceSliceManager) Stop() {
...
+// deviceLookupFunc resolves a (pool, device) tuple to a primary mapping UUID
+// and optional MIG info using the version-specific Lister cache. Versions
+// differ in their attribute layout (v1 stores attributes directly on the
+// device; v1beta1 wraps them under .Basic), so each version supplies its own
+// implementation.
+type deviceLookupFunc func(pool, device string) (string, *DRAMigDeviceInfo)
+
+func makeV1Lookup(lister resourcelistersv1.ResourceSliceLister) deviceLookupFunc {
+	return func(pool, device string) (string, *DRAMigDeviceInfo) {
+		slices, err := lister.List(labels.Everything())
```
rvatkar, line 181: Thanks for splitting out the version-specific lookup here. One concern: GetDeviceInfo now calls lister.List(labels.Everything()) and scans the full cached ResourceSlice set for each lookup. This is a local informer-cache list, not an API-server list, but it is still called once per pod/container/claim-resource from toDeviceToPodsDRA (kubernetes.go:600). That changes the scrape path from the previous O(1) map lookup to O(claims × cached ResourceSlices × devicesPerSlice), and the unfiltered informer means the cost can scale with cluster size even though dcgm-exporter only needs claims for this node.

Could we either scope the informer with WithTweakListOptions where safe, or build/index the relevant ResourceSlice data once per scrape/manager instead of listing and scanning inside every GetDeviceInfo call?

internal/pkg/transformation/dra.go
```diff
@@ -79,100 +152,123 @@ func (m *DRAResourceSliceManager) Stop() {
...
-
-	m.mu.Lock()
-	defer m.mu.Unlock()
-
-	for _, dev := range slice.Spec.Devices {
-		if dev.Basic == nil || dev.Basic.Attributes == nil {
-			continue
+func makeV1beta1Lookup(lister resourcelistersv1beta1.ResourceSliceLister) deviceLookupFunc {
+	return func(pool, device string) (string, *DRAMigDeviceInfo) {
+		slices, err := lister.List(labels.Everything())
```
rvatkar, line 211: Same as https://github.com/NVIDIA/dcgm-exporter/pull/664/changes#r3352651411

internal/pkg/transformation/dra.go
```diff
@@ -22,54 +22,127 @@ import (
...
 	m.cancelContext = cancel
 	factory.Start(ctx.Done())
 
-	if !cache.WaitForCacheSync(ctx.Done(), informer.HasSynced) {
+	// Bound the initial sync. Discovery may say the API is served, but RBAC
+	// can still deny list/watch on resourceslices, in which case HasSynced
+	// would never become true and startup would hang.
+	syncCtx, syncCancel := context.WithTimeout(ctx, resourceSliceCacheSyncTimeout)
+	defer syncCancel()
+	if !cache.WaitForCacheSync(syncCtx.Done(), hasSynced) {
```
rvatkar, line 140: Good call adding the bounded sync timeout. I think we need one cleanup step on the failure path: after cancel(), this should also call factory.Shutdown(). cancel() signals the informer stop channel, but Shutdown() is what waits for the informer goroutines to exit. Stop() does this at :157, but the caller gets nil, err here and never has a manager instance to call Stop() on. The timeout makes this path much more reachable on RBAC-denied clusters.

Could we call m.factory.Shutdown() after cancel() before returning the sync failure error?

internal/pkg/transformation/dra.go
```diff
@@ -79,100 +152,123 @@ func (m *DRAResourceSliceManager) Stop() {
...
+		if err != nil {
+			slog.Debug("listing v1 ResourceSlices failed", "error", err)
+			return "", nil
+		}
+		for _, s := range slices {
+			if s.Spec.Driver != DRAGPUDriverName || s.Spec.Pool.Name != pool {
+				continue
+			}
+			for _, d := range s.Spec.Devices {
+				if d.Name != device {
```
rvatkar, line 191: Small diagnostic suggestion: this returns on the first d.Name == device within the first matching pool slice. If two slices in the same pool advertise the same device name, for example during a stale/rolling-update window, the selected mapping becomes list-order-dependent.

Not a blocker, but a debug log when multiple matches are present would make this easier to diagnose.

Also goes to line R221

internal/pkg/transformation/dra.go
```diff
@@ -22,54 +22,127 @@ import (
...
+)
+
+func (v resourceSliceAPIVersion) String() string {
+	switch v {
+	case resourceSliceAPIV1:
+		return resourcev1.SchemeGroupVersion.String()
+	case resourceSliceAPIV1beta1:
+		return resourcev1beta1.SchemeGroupVersion.String()
+	default:
+		return "unknown"
```
rvatkar, line 65: Tiny nit: resourceSliceAPIUnknown is rejected before the “Using ResourceSlice API” log, so the “unknown” string is effectively unreachable for that path. Fine to keep as a defensive default; just noting it.

internal/pkg/transformation/kubernetes.go
```diff
@@ -652,6 +664,29 @@ func (p *PodMapper) toDeviceToPodsDRA(devicePods *podresourcesapi.ListPodResourc
...
+// key that DCGM uses for per-MIG-instance metrics. The MIG GPU-instance ID is
+// not exposed as a DRA device attribute today, so we look it up via NVML using
+// the MIG UUID published by the DRA driver. Returns "" when NVML or
+// deviceInfo are unavailable, or when the MIG UUID cannot be resolved; in
+// that case the caller falls back to the parent-UUID join key only.
+func (p *PodMapper) draMIGGPUInstanceIdentifier(migInfo *DRAMigDeviceInfo, deviceInfo deviceinfo.Provider) string {
+	if migInfo == nil || deviceInfo == nil || migInfo.MIGDeviceUUID == "" {
+		return ""
+	}
+	migDevice, err := nvmlprovider.Client().GetMIGDeviceInfoByID(migInfo.MIGDeviceUUID)
```
rvatkar, line 677: The new edge-case coverage looks solid: discovery fallback, nil attributes, unknown types, and cross-driver filtering are all covered. One gap that would be great to close: this NVML-failure branch is not exercised yet. A sibling to TestPodDRAInfo_MIGAddsGPUInstanceMapping could make GetMIGDeviceInfoByID return an error and assert that we fall back to parent-UUID-only mapping, with no <gpu>-<gi> key. That would pin the graceful-degradation behavior this PR is adding.

internal/pkg/transformation/kubernetes.go
```diff
@@ -652,6 +664,29 @@ func (p *PodMapper) toDeviceToPodsDRA(devicePods *podresourcesapi.ListPodResourc
...
+	if err != nil {
+		slog.Warn("DRA: NVML lookup for MIG device failed; per-MIG metrics will not be enriched with pod labels",
+			"migUUID", migInfo.MIGDeviceUUID, "error", err)
+		return ""
+	}
+	if migDevice.GPUInstanceID < 0 {
```
rvatkar, line 683: Small follow-up while adding the negative-path test: could we also cover GPUInstanceID < 0 here? The current tests only use non-negative IDs, so this guard is not exercised yet. It should fall back to parent-UUID-only mapping as well.

guptaNswati:
Addressing the open comments in internal MR. cc @rvatkar

ritupanjwani:
Hello Team,
Just checking if there are any updates on when we can expect DCGM metrics to be available for the DRA plugin. Thanks!

nccurry:

PR #671, opened 2026-06-03
packaging/systemd: log to journald instead of an unrotated logfile

### LINKED ISSUE #668: systemd unit should not define StandardOutput and StandardError

Change

When using systemd, the best output for output is journald, not a log file. But if want to keep the /var/log/dcgm-exporter.log output, please at least provide a logrotate configuration file.

### Feature Description

As a system administrator, I don’t want to have another logfile to monitor.

Just remove the two directives StandardOutput and StandardError.

### Additional context

_No response_

75asu:
Closes #668.

Removes the `StandardOutput`/`StandardError` `append:` directives from the systemd unit so the exporter inherits systemd's default `journal` sink. Output is then queryable via `journalctl -u nvidia-dcgm-exporter`, rotated/size-capped by journald, and structured with unit/PID/priority, instead of writing to an unrotated `/var/log/dcgm-exporter.log` that operators must manage separately. This matches the change requested in the issue.

### DIFF
--- a/packaging/config-files/systemd/nvidia-dcgm-exporter.service
+++ b/packaging/config-files/systemd/nvidia-dcgm-exporter.service
@@ -22,9 +22,6 @@ After=nvidia-dcgm.service
 User=root
 PrivateTmp=false
 
-StandardOutput=append:/var/log/dcgm-exporter.log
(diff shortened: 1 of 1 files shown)


nccurry:

Issue #612, opened 2026-01-13
dcgm-exporter exiting with status 1 and no apparent error message


m3co-code:
### What is the version?

4.4.1-4.6.0

### What happened?

The dcgm-exporter is going into Crashloop Backoff in a Kubernetes environment with exit status 1 and no apparent reason.. It is not happening on all instances but only on some. Interestingly, the pod is crashlooping, but if I manually delete it, it recovers.

I'd be happy about further debugging help, I tried adding debugging env vars but don't get an error log.

```
$ k logs -n kube-system dcgm-exporter-6xn9b
Defaulted container "nvidia-dcgm-container" out of: nvidia-dcgm-container, do-node-agent, dynamic-config (init)
time=2026-01-13T14:23:40.195Z level=DEBUG msg="Debug output is enabled"
time=2026-01-13T14:23:40.195Z level=INFO msg="Starting dcgm-exporter" Version=4.4.1-4.6.0
exit status 1
```

### What did you expect to happen?

The exporter working across nodes.

### What is the GPU model?

h100x1

```
# nvidia-smi
Tue Jan 13 14:37:13 2026
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 575.57.08              Driver Version: 575.57.08      CUDA Version: 12.9     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA H100 80GB HBM3          On  |   00000000:00:09.0 Off |                    0 |
| N/A   40C    P0             77W /  700W |       0MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
```

### What is the environment?

Kubernetes version 1.33.

### How did you deploy the dcgm-exporter and what is the configuration?

With a daemonset that was created based on the helm charts.

```
apiVersion: apps/v1
[...]

PR #679, opened 2026-06-16
helm: allow overriding probe timeoutSeconds/failureThreshold

ChenZiHong-Gavin:
## What

Expose `timeoutSeconds` and `failureThreshold` (plus `readinessProbe.periodSeconds`)
as optional values on the liveness and readiness probes in the DaemonSet chart.

## Why

The chart currently only templatizes `initialDelaySeconds` and (for liveness)
`periodSeconds`. It never sets `timeoutSeconds` or `failureThreshold`, so the
kubelet defaults apply: `timeoutSeconds: 1`, `failureThreshold: 3`.

The liveness probe hits `/health` over HTTP. On busy nodes that endpoint can take
longer than 1s to respond; three consecutive >1s responses trip the default
`failureThreshold: 3` and the kubelet kills an otherwise-healthy container
(graceful exit 0, not OOM), which then restarts and can repeat — a false-kill /
crashloop. Today there is no values-based way to loosen the probe; the only
options are vendoring the chart or post-rendering the manifest.

This adds the missing knobs so operators can set e.g. `timeoutSeconds: 5` /
`failureThreshold: 5` (see #612 for a report of the same symptom).

## How

The new fields are rendered with `{{- with ... }}`, so when they are unset the
template emits nothing and the kubelet defaults continue to apply. **Existing
installations see no change** unless they explicitly set the new values.

## Validation

- `helm template` with no overrides: rendered probes are byte-for-byte unchanged
  (no `timeoutSeconds` / `failureThreshold` emitted).
- `helm template` with the new values set: fields render correctly on both the
  `httpGet` and `basicAuth` `tcpSocket` probe branches.
- `helm lint`: passes.

### DIFF
--- a/deployment/templates/daemonset.yaml
+++ b/deployment/templates/daemonset.yaml
@@ -198,6 +198,12 @@ spec:
           {{- end }}
           initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
           periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
+          {{- with .Values.livenessProbe.timeoutSeconds }}
+          timeoutSeconds: {{ . }}
+          {{- end }}
+          {{- with .Values.livenessProbe.failureThreshold }}
+          failureThreshold: {{ . }}
+          {{- end }}
         readinessProbe:
           {{- if not $.Values.basicAuth.users }}
           httpGet:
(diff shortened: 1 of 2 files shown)


ChenZiHong-Gavin:
@nccurry @daphne2337 @jay-mckay @glowkey @guptaNswati Please review this PR. Thank you.

nccurry:
This should be fixed in 4.8.4. Closing. Please re-open or open a new issue or PR if you find issues.

PR #687, opened 2026-06-20
server: thread r.Context() through render (golangci-lint contextcheck)

randomizedcoder:
What: pass the HTTP request context through `render()` and use it for logging + cancellation.
Why: Prometheus scrapers cancel on timeout; today the daemon keeps rendering after cancellation, wasting CPU and adding shutdown latency. Also fixes two `slog.LogAttrs` calls that currently use `context.Background()` as a placeholder.
How: `Metrics(w, r)` reads `r.Context()`; `render(ctx, w, mg)`; `ctx.Err()` check between metric groups.

Found by: golangci-lint 2.12.2, linter **contextcheck** — *function calls a context-using callee without threading the caller's context*.

Follow-up: thread `ctx` into `transformation.Process` and `rendermetrics.RenderGroup` (separate PR; needs interface changes).

Behaviour change to note in release notes: a scrape cancelled mid-render now returns a truncated body. Some monitoring setups may surface this as a Prometheus parse error.

### DIFF
--- a/internal/pkg/server/server.go
+++ b/internal/pkg/server/server.go
@@ -246,7 +246,8 @@ func (s *MetricsServer) fatal() {
 	os.Exit(1)
 }
 
-func (s *MetricsServer) Metrics(w http.ResponseWriter, _ *http.Request) {
+func (s *MetricsServer) Metrics(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


nccurry:

PR #697, opened 2026-06-21
Use local hostname when remote hostengine is on loopback

### LINKED ISSUE #572: Changing the NV-Hostengine port will result in `Hostname` label to always be `localhost`

Change

The DCGM exporter changed the logic of setting the `Hostname` label in v4. The exporter sets the label to use the hostname provided through the argument `-r`. See [this](https://github.com/NVIDIA/dcgm-exporter/blob/13ad457b344272c3252a7d5be5bcb3fbf6e1c65f/internal/pkg/hostname/hostname.go#L34) and [this](https://github.com/NVIDIA/dcgm-exporter/blob/13ad457b344272c3252a7d5be5bcb3fbf6e1c65f/pkg/cmd/app.go#L714).

Changing the default port of `nv-hostengine`, let's say `localhost:5556`, the label will be set to `localhost`. That means that it will not be possible to filter the metrics by host in Grafana dashboards.

### Feature Description

As a user, I still want to be able to change the default port on localhost and still see the correct hostname set on the label.

A possible solution would be to additionally check if the `config.RemoteHEInfo` is set to `localhost` or `127.0.0.1` and use the `os.Hostname()` rather than relying on the value of the config.

### Additional context

https://github.com/canonical/dcgm-snap/issues/71

skartikey:
In v4 the `hostname` label is derived from the host portion of the remote hostengine address passed via `-r` / `DCGM_REMOTE_HOSTENGINE_INFO`. When the hostengine runs on the local node but on a non-default port (for example
  `localhost:5556` or `127.0.0.1:5556`), the label ends up being `localhost` / `127.0.0.1`. That makes it impossible to distinguish or filter metrics by node in Grafana dashboards.

  This change detects loopback remote addresses and falls back to the local hostname, matching the behavior the exporter already uses for the Kubernetes path.

  Loopback detection covers:

  - `localhost` (case-insensitive)
  - IPv4 loopback `127.0.0.0/8`
  - IPv6 loopback `::1`
  - bracketed and port-suffixed forms (`[::1]:5555`, `127.0.0.1:5556`, ...)

  Non-loopback hosts (`example.com`, public IPv4/IPv6, the unspecified address `::`) are unaffected and still used verbatim. The fallback honors `NODE_NAME` first, then `os.Hostname()`, identical to the existing local path.

Closes #572

### DIFF
--- a/internal/pkg/hostname/hostname.go
+++ b/internal/pkg/hostname/hostname.go
@@ -18,6 +18,7 @@ package hostname
 
 import (
 	"net"
+	"strings"
 
 	"github.com/NVIDIA/dcgm-exporter/internal/pkg/appconfig"
 	osinterface "github.com/NVIDIA/dcgm-exporter/internal/pkg/os"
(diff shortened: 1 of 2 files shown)


nccurry:

PR #665, opened 2026-05-30
feat(helm): make automountServiceAccountToken configurable

75asu:
Resolves #624

- Adds `serviceAccount.automountServiceAccountToken` to expose the field as an explicit override.
- Backward compatible: when unset, falls back to the existing derivation from `kubernetes.enablePodLabels`/`enablePodUID` with `rbac.create`, or `kubernetesDRA.enabled`.
- Applied to both the DaemonSet pod spec and the ServiceAccount for consistency.
- Verified with `helm template` for default, explicit `true`, explicit `false`, and the DRA-enabled derivation path.

PR #698, opened 2026-06-21
Fall back to metrics file instead of crashing on ConfigMap read failure

### LINKED ISSUE #624: DCGM Exporter DaemonSet crashes in Kubernetes due to missing ServiceAccount token (/var/run/secrets/.../token)
### **Summary**

dcgm-exporter fails to run reliably as a DaemonSet in Kubernetes when DCGM_EXPORTER_KUBERNETES is enabled.
The container crashes with:`open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
`
Even when a ServiceAccount exists and is referenced, the token is not mounted or is ignored, causing the exporter to crash repeatedly.

### **Environment**

**Kubernetes version:** v1.28.15
**OS:** Ubuntu 22.04
**GPU nodes:** NVIDIA DGX A100 (SXM4 40GB)
**Driver:** 535.xx
**CUDA:** 12.2
**MIG mode:** Enabled (mixed)
**Container runtime:** containerd
**Helm chart:** dcgm-exporter
**Image:** `nvcr.io/nvidia/k8s/dcgm-exporter:4.4.2-4.7.1-ubuntu22.04`

### **Expected Behavior**

The dcgm-exporter DaemonSet should:

  - Start successfully on all GPU nodes
  - Collect GPU / MIG metrics
  - Optionally map GPU usage to Kubernetes pods when `DCGM_EXPORTER_KUBERNETES=true`

### **Actual Behavior**

  - Pods enter CrashLoopBackOff
  - Logs consistently show: `ERROR msg="open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory"`
  
  - This happens even when:
          - A ServiceAccount exists
          - The ServiceAccount is explicitly set in Helm values
          - `automountServiceAccountToken` is enabled
       

### **Pod Logs**
```
time=2026-01-29T07:46:24Z level=INFO msg="Starting dcgm-exporter"
time=2026-01-29T07:46:31Z level=INFO msg="DCGM successfully initialized!"
time=2026-01-29T07:46:32Z level=INFO msg="Collecting DCP Metrics"
time=2026-01-29T07:46:32Z level=ERROR msg="open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory"

```

### Helm Values used

```
arguments:
  - "-c=1000"
  - "--collectors=nvml,dcgm"

extraEnv:
  - name: DCGM_EXPORTER_KUBERNETES
    value: "true"

nodeSelector:
  nvidia.com/gpu.present: "true"

resources:
  requests:
    cpu: "250m"
    memory: "512Mi"
  limits:
    cpu: "1"
    memory: "1Gi"

serviceAccount:
  create: false
  name: dcgm-exporter

automountServiceAccountToken: true
```

### Additional Observations

  - nvidia-dcgm.service is running correctly on the host:
    ```
    systemctl status nvidia-dcgm
    Active: active (running)
    
    ```
  - `/var/lib/kubelet/pod-resources` is mounted correctly
  - GPU labels `(nvidia.com/gpu.present=true)` are present
  - Node has no taints
  - Issue occurs on multiple nodes
  - Disabling Kubernetes mode:

     `DCGM_EXPORTER_KUBERNETES=false`

     avoids the token error, but then pod-level metrics are unavailable

### Questions

  - How to make this pod up and running?
  - Is `DCGM_EXPORTER_KUBERNETES=true` strictly required for dcgm-exporter DaemonSet?
  - Is there a supported way to run dcgm-exporter in Kubernetes without requiring a ServiceAccount token?
  - Is Kubernetes v1.28 officially supported for pod-level GPU metrics?

### Thank you

Thanks for maintaining dcgm-exporter — any guidance on the correct configuration or confirmation of a bug would be appreciated.

@nvidia
@kubernetes
@helm
@prometheus
@grafana

skartikey:
When a counters ConfigMap is configured (the default for the Helm chart), `counters.GetCounterSet` creates an in-cluster Kubernetes client and called `os.Exit(1)` if either the client creation or the ConfigMap read failed.

In a pod where the ServiceAccount token is not mounted, this turns a recoverable condition into a `CrashLoopBackOff`, with only the raw error in the logs:

```
level=INFO msg="Collecting DCP Metrics"
level=ERROR msg="open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory"
```

The exit happens even though `GetCounterSet` already has a fallback-to-metrics-file path: the `os.Exit(1)` short-circuits it before it can run.

**Behavior change**

- ConfigMap configured but no kube client (e.g. no SA token): warn + fall back to the metrics file, exporter starts and serves metrics (using default counters) instead of crash-looping.
- ConfigMap configured but unreadable/missing: same graceful fallback.
- ConfigMap available: unchanged, counters are read from the ConfigMap.
- No ConfigMap configured: unchanged.

Closes #624

### DIFF
--- a/internal/pkg/counters/counter_config.go
+++ b/internal/pkg/counters/counter_config.go
@@ -29,7 +29,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/NVIDIA/dcgm-exporter/internal/pkg/appconfig"
-	"github.com/NVIDIA/dcgm-exporter/internal/pkg/kubeclient"
 )
 
 func GetCounterSet(ctx context.Context, c *appconfig.Config) (*CounterSet, error) {
@@ -42,15 +41,18 @@ func GetCounterSet(ctx context.Context, c *appconfig.Config) (*CounterSet, error
 
 	if c.ConfigMapData != undefinedConfigMapData {
 		var client kubernetes.Interface
-		client, err = kubeclient.GetKubeClient()
+		client, err = getKubeClient()
 		if err != nil {
-			slog.Error(err.Error())
-			os.Exit(1)
-		}
-		records, err = readConfigMap(ctx, client, c)
-		if err != nil {
-			slog.Error(err.Error())
-			os.Exit(1)
+			slog.Warn("Failed to create Kubernetes client to read the counters ConfigMap; "+
+				"falling back to the metrics file. In Kubernetes this usually means the pod has no "+
+				"mounted ServiceAccount token (/var/run/secrets/kubernetes.io/serviceaccount/token); "+
+				"ensure a ServiceAccount is configured and automountServiceAccountToken is enabled.",
+				"configMap", c.ConfigMapData,
+				"error", err)
+		} else if records, err = readConfigMap(ctx, client, c); err != nil {
+			slog.Warn("Failed to read the counters ConfigMap; falling back to the metrics file.",
+				"configMap", c.ConfigMapData,
+				"error", err)
 		}
 	} else {
 		err = fmt.Errorf("no configmap data specified")
--- a/internal/pkg/counters/counter_config_test.go
+++ b/internal/pkg/counters/counter_config_test.go
@@ -18,6 +18,7 @@ package counters
 
 import (
 	"context"
+	"errors"
 	stdos "os"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


skartikey:
Note: this is complementary to #665, not a replacement. #665 fixes the root cause in the Helm chart (mounting the ServiceAccount token). This PR is defense-in-depth in the binary: regardless of why the token is absent, `GetCounterSet` shouldn't `os.Exit(1)` and crash-loop. It already has a fallback-to-metrics-file path that the hard exit was short-circuiting. Happy to switch to a clear fail-fast error instead of the silent fallback if you'd prefer that a custom ConfigMap never be quietly bypassed.

rvatkar:
We decided not to add automatic fallback. When a user explicitly configures a metric source, failure to read it indicates an infrastructure or configuration problem and dcgm-exporter should fail clearly. Falling back could hide that problem and serve an unexpected metric set.
Issue [#624](https://github.com/NVIDIA/dcgm-exporter/issues/624) was addressed separately in release 4.6.0-4.8.3. Closing this fallback request.

PR #699, opened 2026-06-21
Drop direct dependency on archived github.com/pkg/errors

### LINKED ISSUE #526: Suggest to remove EOL dependency: github.com/pkg/errors
**What would you like to be added**:

Replace https://github.com/pkg/errors with Golang native error handling.

**Why is this needed**:

I see this project is still using the package github.com/pkg/errors:
https://github.com/NVIDIA/dcgm-exporter/blob/8fd69339fe2138459d3618508c3ebb1f43b51f79/go.mod#L16

Note that this package was archived by the owner in 2021. 

Does it make sense to get rid of it?

skartikey:
github.com/pkg/errors was archived by its author in 2021. It was used in a single place (tests/e2e/internal/framework/kube.go) via errors.Wrap. Replace it with the standard library equivalent fmt.Errorf("...: %w", err), which the same file already uses elsewhere, and run go mod tidy.

The package now only remains as an indirect dependency pulled in transitively by helm.sh/helm/v3 (via mittwald/go-helm-client), which is outside this repository's control.

Closes #526

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/mittwald/go-helm-client v0.12.19
 	github.com/onsi/ginkgo/v2 v2.28.1
 	github.com/onsi/gomega v1.39.1
-	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.67.5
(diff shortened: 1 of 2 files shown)


nccurry:

PR #701, opened 2026-06-27
fix: skip GPM/DCP metrics on fractional vGPU guests

### LINKED ISSUE #661: DCGM Exporter doesn't work with fractional vGPU shapes (nvidia-rtx-pro-6000)
### What is the version?

4.4.1-4.6.0

### What happened?

I have GKE cluster with DCGM Exporter version `4.4.1-4.6.0`. I saw following error log:
```
msg="Got unexpected return -14 from m_gpmManager.GetLatestSample [/workspaces/dcgm-exporter/dcgmlib/src/DcgmCacheManager.cpp:13088] [DcgmCacheManager::BufferOrCacheLatestGpuValue]" dcgm_level=ERROR"
```
from the node for G4 nodes that run fractional vGPU shapes (g4-standard-24, g4-standard-12, g4-standard-6), while they work as expected on G4 machines that run >=1 GPU, so G4 machines >= g4-standard-48.

(G4 machine: https://docs.cloud.google.com/compute/docs/accelerator-optimized-machines#g4-series)

Only node with full GPU has the metric 

Image

### What did you expect to happen?

I can see DCGM metrics exported from G4 node with fractional vGPU shapes ( < 1 GPU).

### What is the GPU model?

nvidia-rtx-pro-6000

### What is the environment?

DCGM-Exporter runs on GKE cluster as a daemonset.

### How did you deploy the dcgm-exporter and what is the configuration?

Run it as a daemonset, part of the set-up:
```
      containers:
      - args:
        - --enable-dcgm-log
        - --dcgm-log-level
        - ERROR
        - name: DCGM_EXPORTER_KUBERNETES_GPU_ID_TYPE
          value: device-name
        - name: DCGM_EXPORTER_KUBERNETES
          value: "true"
        - name: DISABLE_STARTUP_VALIDATE
          value: "true"
```

### How to reproduce the issue?

_No response_

### Anything else we need to know?

I wonder if I need to make changes to configuration in DCGM-Exporter to make it work, or it indeed doesn't work on nodes with partial nvidia-rtx-pro-6000.

I also tried [4.5.2-4.8.1](https://hub.docker.com/layers/nvidia/dcgm-exporter/4.5.2-4.8.1-distroless/images/sha256-287d540dc60489ab5dce907dddaba0b5d1970b8c7c14f1aa8b624dbee7b4405c) but still see the same error.

kshitizlohia1994:
## Problem

On GKE G4 nodes with fractional GPU allocations (g4-standard-6, g4-standard-12, g4-standard-24), dcgm-exporter emits repeated errors:

```
Got unexpected return -14 from m_gpmManager.GetLatestSample
```

and exports no profiling metrics. Full GPU shapes (g4-standard-48+) are unaffected.

The root cause is that fractional vGPU guests run under `GPU_VIRTUALIZATION_MODE_VGPU`, which prevents GPM (GPU Performance Monitoring) from collecting samples. The exporter attempted to collect DCP/profiling metrics unconditionally, resulting in a continuous error loop.

Fixes #661

## Solution

Introduce `internal/pkg/profiling` with `ValidateGPMSupport()`, which runs before DCP metric collection on each reload cycle and:

1. Queries each GPU's virtualization mode via NVML `GetVirtualizationMode` — immediately returns unsupported for vGPU guest modes.
2. Queries `GpmQueryDeviceSupport` to check hardware-level GPM capability.
3. Performs a live two-sample GPM probe to confirm metrics can actually be retrieved at runtime.

When any GPU fails validation, DCP/profiling collection is disabled for that cycle while all standard DCGM metrics continue to be exported normally.

## Test plan

- [x] Unit tests for `VirtualizationModeBlocksGPM` covering all virtualization mode variants
- [ ] Verify on a full GPU shape (g4-standard-48+): DCP metrics continue to be collected
- [ ] Verify on a fractional vGPU shape (g4-standard-6/12/24): no GPM errors, standard metrics still exported

### DIFF
--- a/internal/pkg/profiling/gpm.go
+++ b/internal/pkg/profiling/gpm.go
@@ -1,0 +1,156 @@
+/*
+ * Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package profiling validates runtime support for DCGM profiling (DCP/GPM) metrics.
+package profiling
+
+import (
+	"fmt"
+	"log/slog"
+	"time"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+
+	"github.com/NVIDIA/dcgm-exporter/internal/pkg/dcgmprovider"
+)
+
+const gpmProbeSampleInterval = 500 * time.Millisecond
+
+// VirtualizationModeBlocksGPM reports whether the NVML virtualization mode prevents
+// collection of GPM-backed DCGM_FI_PROF_* metrics. Fractional vGPU guests (for example
+// GKE G4 g4-standard-6/12/24 shapes) run in vGPU mode and cannot provide GPM samples.
+func VirtualizationModeBlocksGPM(mode nvml.GpuVirtualizationMode) bool {
+	switch mode {
+	case nvml.GPU_VIRTUALIZATION_MODE_VGPU, nvml.GPU_VIRTUALIZATION_MODE_HOST_VGPU:
+		return true
+	default:
+		return false
+	}
+}
+
+// ValidateGPMSupport checks whether GPM profiling metrics can be collected on all
+// visible GPUs. When validation fails, callers should disable DCP metric collection
+// while continuing to export standard NVML-backed DCGM metrics.
+func ValidateGPMSupport() (bool, string) {
+	gpuCount, err := dcgmprovider.Client().GetAllDeviceCount()
+	if err != nil {
+		return false, fmt.Sprintf("failed to get GPU count: %v", err)
+	}
+	if gpuCount == 0 {
+		return false, "no GPUs found"
+	}
+
+	ret := nvml.Init()
+	if ret != nvml.SUCCESS {
+		slog.Warn("Skipping GPM validation because NVML could not be initialized",
+			slog.String("error", nvml.ErrorString(ret)))
+		return true, ""
+	}
+	defer nvml.Shutdown()
+
+	for gpuID := uint(0); gpuID < gpuCount; gpuID++ {
+		gpuInfo, err := dcgmprovider.Client().GetDeviceInfo(gpuID)
+		if err != nil {
+			return false, fmt.Sprintf("failed to get device info for GPU %d: %v", gpuID, err)
+		}
+
+		device, ret := nvml.DeviceGetHandleByUUID(gpuInfo.UUID)
+		if ret != nvml.SUCCESS {
+			return false, fmt.Sprintf("failed to get NVML handle for GPU %s: %s", gpuInfo.UUID, nvml.ErrorString(ret))
+		}
+
+		if mode, ret := device.GetVirtualizationMode(); ret == nvml.SUCCESS {
+			if VirtualizationModeBlocksGPM(mode) {
+				return false, fmt.Sprintf(
+					"GPU %s is in vGPU virtualization mode; GPM profiling metrics are not available",
+					gpuInfo.UUID,
+				)
+			}
+		} else {
+			slog.Debug("Could not query GPU virtualization mode",
+				slog.String("gpu_uuid", gpuInfo.UUID),
+				slog.String("error", nvml.ErrorString(ret)))
+		}
+
+		gpmSupport, ret := device.GpmQueryDeviceSupport()
+		if ret != nvml.SUCCESS {
+			return false, fmt.Sprintf("GPU %s GPM support query failed: %s", gpuInfo.UUID, nvml.ErrorString(ret))
+		}
+		if gpmSupport.IsSupportedDevice == 0 {
+			return false, fmt.Sprintf("GPU %s does not support GPM", gpuInfo.UUID)
+		}
+
+		if !probeGPMSample(device) {
+			return false, fmt.Sprintf("GPU %s failed GPM sample probe", gpuInfo.UUID)
+		}
+	}
+
+	return true, ""
+}
+
+func probeGPMSample(device nvml.Device) bool {
+	sample1, ret := nvml.GpmSampleAlloc()
+	if ret != nvml.SUCCESS {
+		slog.Debug("GPM sample allocation failed", slog.String("error", nvml.ErrorString(ret)))
+		return false
+	}
+	defer func() {
+		if freeRet := nvml.GpmSampleFree(sample1); freeRet != nvml.SUCCESS {
+			slog.Debug("GPM sample free failed", slog.String("error", nvml.ErrorString(freeRet)))
+		}
+	}()
+
+	sample2, ret := nvml.GpmSampleAlloc()
+	if ret != nvml.SUCCESS {
+		slog.Debug("GPM sample allocation failed", slog.String("error", nvml.ErrorString(ret)))
+		return false
+	}
+	defer func() {
+		if freeRet := nvml.GpmSampleFree(sample2); freeRet != nvml.SUCCESS {
+			slog.Debug("GPM sample free failed", slog.String("error", nvml.ErrorString(freeRet)))
+		}
+	}()
+
+	if ret := device.GpmSampleGet(sample1); ret != nvml.SUCCESS {
+		slog.Debug("First GPM sample failed", slog.String("error", nvml.ErrorString(ret)))
+		return false
+	}
+
+	time.Sleep(gpmProbeSampleInterval)
+
+	if ret := device.GpmSampleGet(sample2); ret != nvml.SUCCESS {
+		slog.Debug("Second GPM sample failed", slog.String("error", nvml.ErrorString(ret)))
+		return false
+	}
+
+	metricsGet := nvml.GpmMetricsGetType{
+		Sample1:    sample1,
+		Sample2:    sample2,
+		NumMetrics: 1,
+	}
+	metricsGet.Metrics[0].MetricId = uint32(nvml.GPM_METRIC_GRAPHICS_UTIL)
+
+	if ret := nvml.GpmMetricsGet(&metricsGet); ret != nvml.SUCCESS {
+		slog.Debug("GPM metrics query failed", slog.String("error", nvml.ErrorString(ret)))
+		return false
+	}
+	if metricsGet.Metrics[0].NvmlReturn != uint32(nvml.SUCCESS) {
+		slog.Debug("GPM metric returned NVML error", slog.Uint64("nvml_return", uint64(metricsGet.Metrics[0].NvmlReturn)))
+		return false
+	}
+
+	return true
+}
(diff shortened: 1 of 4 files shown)


kshitizlohia1994:
@nccurry  can you please review this pr when you get a chance?

internal/pkg/profiling/gpm.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+const gpmProbeSampleInterval = 500 * time.Millisecond
+
+// VirtualizationModeBlocksGPM reports whether the NVML virtualization mode prevents
+// collection of GPM-backed DCGM_FI_PROF_* metrics. Fractional vGPU guests (for example
+// GKE G4 g4-standard-6/12/24 shapes) run in vGPU mode and cannot provide GPM samples.
+func VirtualizationModeBlocksGPM(mode nvml.GpuVirtualizationMode) bool {
+	switch mode {
+	case nvml.GPU_VIRTUALIZATION_MODE_VGPU, nvml.GPU_VIRTUALIZATION_MODE_HOST_VGPU:
```
rvatkar: Could we avoid rejecting every VGPU/HOST_VGPU mode here? NVML supports GPM for some full-framebuffer MIG-backed vGPU profiles. Returning before the capability query and probe could disable a supported configuration. Could we rely on those capability checks, or narrow this condition to positively identified fractional profiles?

internal/pkg/profiling/gpm.go
```diff
@@ -0,0 +1,156 @@
...
+					gpuInfo.UUID,
+				)
+			}
+		} else {
+			slog.Debug("Could not query GPU virtualization mode",
+				slog.String("gpu_uuid", gpuInfo.UUID),
+				slog.String("error", nvml.ErrorString(ret)))
+		}
+
+		gpmSupport, ret := device.GpmQueryDeviceSupport()
```
rvatkar: Could we avoid treating “NVML GPM unsupported” as “DCGM profiling unsupported”? DCGM explicitly routes GPM-capable devices through m_gpmManager and non-GPM profiling fields through its [Profiling Module](https://github.com/NVIDIA/DCGM/blob/72fa3feaa67d716a75323a8f47c34ff3ee73f824/dcgmlib/src/DcgmCacheManager.cpp#L3617-L3632) , while NVML documents [GPM sampling](https://github.com/NVIDIA/DCGM/blob/72fa3feaa67d716a75323a8f47c34ff3ee73f824/sdk/nvidia/nvml/dcgm_nvml.h#L12345-L12402) as Hopper+. 
As written, an A100 could return supported DCGM metric groups and then be rejected here. Could this probe be limited to the GPM-backed path while retaining DCGM’s result for non-GPM GPUs?

internal/pkg/profiling/gpm.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	defer nvml.Shutdown()
+
+	for gpuID := uint(0); gpuID < gpuCount; gpuID++ {
+		gpuInfo, err := dcgmprovider.Client().GetDeviceInfo(gpuID)
+		if err != nil {
+			return false, fmt.Sprintf("failed to get device info for GPU %d: %v", gpuID, err)
+		}
+
+		device, ret := nvml.DeviceGetHandleByUUID(gpuInfo.UUID)
```
rvatkar: How should this validation behave with `--remote-hostengine-info`? The UUID comes from the remote DCGM host, but this lookup uses local NVML. If local NVML is initialized but does not contain that UUID, otherwise-supported remote profiling will be disabled. Could remote mode retain DCGM’s capability result instead of performing this local lookup?

pkg/cmd/app.go
```diff
@@ -979,6 +980,15 @@ func queryDCPMetrics(config *appconfig.Config, reloadID uint64) {
 		return
 	}
 
+	if supported, reason := profiling.ValidateGPMSupport(); !supported {
+		config.CollectDCP = false
```
rvatkar, line 987: Is disabling DCP globally when any visible GPU fails validation the intended behavior? On a heterogeneous node, one unsupported or temporarily unhealthy GPU would remove profiling from all supported GPUs. Could we either make the decision per device/group or explicitly document and test this all-or-nothing policy?

internal/pkg/profiling/gpm_test.go
```diff
@@ -0,0 +1,59 @@
...
+package profiling
+
+import (
+	"testing"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestVirtualizationModeBlocksGPM(t *testing.T) {
```
rvatkar: Could we add coverage for ValidateGPMSupport and its integration with queryDCPMetrics? At minimum, it would be useful to test pre-Hopper profiling, a supported vGPU, remote hostengine, mixed capabilities, the affected fractional-vGPU failure, and preservation of standard metrics when profiling is disabled.

internal/pkg/profiling/gpm.go
kshitizlohia1994: good point. dropped the virtualization-mode check entirely and now rely on a live gpm sample probe, so supported mig-backed vgpu profiles pass instead of being rejected.

internal/pkg/profiling/gpm.go
kshitizlohia1994: gpus that don't support gpm (e.g. pre-hopper a100) are now skipped, not rejected; their DCGM_FI_PROF_* fields keep flowing through the dcgm profiling module.

internal/pkg/profiling/gpm.go
kshitizlohia1994: when --remote-hostengine-info is set we short-circuit and trust dcgm's capability result, since local nvml can't resolve remote gpu uuids; a missing local handle maps to "unknown" and never disables profiling.

pkg/cmd/app.go
kshitizlohia1994: changed to a per-gpu decision. profiling is only disabled globally when every gpm-capable gpu fails the probe and no other gpu can serve profiling, so one broken/unhealthy gpu no longer removes profiling from supported gpus.

internal/pkg/profiling/gpm_test.go
kshitizlohia1994: added. table tests for ValidateGPMSupport (pre-hopper, healthy gpm, fractional-vgpu failure, mixed, remote hostengine, unknown) plus queryDCPMetrics integration covering CollectDCP/MetricGroups on keep vs disable.

internal/pkg/profiling/gpm.go
```diff
@@ -0,0 +1,267 @@
...
+// decideDCPFromGPMStatuses implements the per-GPU DCP policy. Profiling is disabled
+// globally only when at least one GPM-capable GPU is broken and no GPU (healthy GPM,
+// non-GPM/DCP-module, or not-yet-determined) can serve profiling metrics.
+func decideDCPFromGPMStatuses(statuses []gpuGPMStatus) (bool, string) {
+	healthy := countStatus(statuses, gpmHealthy)
+	broken := countStatus(statuses, gpmBroken)
+	notApplicable := countStatus(statuses, gpmNotApplicable)
+	unknown := countStatus(statuses, gpmUnknown)
+
+	if broken > 0 && healthy == 0 && notApplicable == 0 && unknown == 0 {
```
rvatkar, line 194: Thanks for updating this. I think the mixed-GPU case still leaves the original error path active. The function returns only a global Boolean, so {healthy, broken} keeps profiling enabled for the shared GPU watch group. DCGM will therefore still call GPM for the broken GPU and emit the GetLatestSample error. Could we either exclude broken GPU IDs from profiling watches or use an explicitly documented all-or-nothing policy?

internal/pkg/profiling/gpm_test.go
```diff
@@ -0,0 +1,271 @@
...
+				m.EXPECT().GetDeviceInfo(uint(1)).Return(dcgm.Device{UUID: "GPU-1"}, nil)
+			},
+			checker: &fakeChecker{initOK: true, statuses: map[string]gpuGPMStatus{
+				"GPU-0": gpmBroken,
+				"GPU-1": gpmBroken,
+			}},
+			wantDisable: true,
+		},
+		{
+			name: "mixed healthy and broken GPM keeps profiling for healthy GPU",
```
rvatkar, line 230: Thanks for adding these tests. This case only asserts that profiling remains globally enabled; it does not prove that the broken GPU is excluded from profiling watches, so it passes with the current incomplete behavior. Could we add a test verifying that healthy GPUs retain profiling, broken GPUs are not watched for profiling fields, and standard metrics remain available when profiling is disabled?

internal/pkg/profiling/gpm.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	if !deviceChecker.init() {
+		// NVML unavailable; keep existing behavior rather than disabling.
+		return false, ""
+	}
+	defer deviceChecker.shutdown()
+
+	statuses := make([]gpuGPMStatus, 0, gpuCount)
+	for gpuID := uint(0); gpuID < gpuCount; gpuID++ {
```
rvatkar, line 109: Thanks for handling mixed capabilities. Could validation also respect config.GPUDeviceOptions? It currently evaluates every visible GPU, including unmonitored ones. For example, an unmonitored healthy GPU could keep profiling enabled for the only monitored—but broken—GPU. Could we base the decision on the entities that will actually be watched?

internal/pkg/profiling/gpm.go
```diff
@@ -0,0 +1,267 @@
...
+			slog.String("error", nvml.ErrorString(ret)))
+		return gpmUnknown
+	}
+	if gpmSupport.IsSupportedDevice == 0 {
+		// GPU does not use GPM for profiling; the DCGM profiling module serves its
+		// DCGM_FI_PROF_* fields, so GPM sample availability is irrelevant here.
+		return gpmNotApplicable
+	}
+
+	if !probeGPMSample(device) {
```
rvatkar, line 179: Could we account for MIG entity sampling here? This probes the physical device with GpmSampleGet, while DCGM samples FE_GPU_I entities using GpmMigSampleGet with the GPU-instance ID. A whole-device probe therefore does not validate the path DCGM will use. Could we probe the actual monitored entity type, or conservatively skip MIG validation until that path is supported and tested?

internal/pkg/profiling/gpm.go
```diff
@@ -0,0 +1,267 @@
...
+			slog.Debug("GPM sample free failed", slog.String("error", nvml.ErrorString(freeRet)))
+		}
+	}()
+
+	if ret := device.GpmSampleGet(sample1); ret != nvml.SUCCESS {
+		slog.Debug("First GPM sample failed", slog.String("error", nvml.ErrorString(ret)))
+		return false
+	}
+
+	time.Sleep(gpmProbeSampleInterval)
```
rvatkar, line 243: Could we avoid sleeping once per GPM-capable GPU? Since validation is serial and runs synchronously during startup/topology reload, eight GPUs add at least four seconds of delay. One option would be to collect the first sample for every GPU, sleep once, then collect the second samples; bounded concurrency could also work.

internal/pkg/profiling/gpm_test.go
```diff
@@ -0,0 +1,271 @@
...
+	statuses := []gpuGPMStatus{gpmHealthy, gpmBroken, gpmBroken, gpmNotApplicable, gpmUnknown}
+
+	assert.Equal(t, 1, countStatus(statuses, gpmHealthy))
+	assert.Equal(t, 2, countStatus(statuses, gpmBroken))
+	assert.Equal(t, 1, countStatus(statuses, gpmNotApplicable))
+	assert.Equal(t, 1, countStatus(statuses, gpmUnknown))
+	assert.Equal(t, 0, countStatus(nil, gpmBroken))
+}
+
+// fakeChecker is a test double for gpmDeviceChecker that returns pre-programmed
```
rvatkar, line 108: Thanks for adding the checker seam. Could we move the seam closer to the NVML calls? The current fake returns a final status, so the production statusForUUID and probeGPMSample paths remain untested.
It would be useful to cover capability-query errors, both sample failures, metric-level errors, and sample cleanup on each failure path.

PR #730, opened 2026-09-05
fix: deduplicate HPC job IDs before metric enrichment

git-jxj:
Repeated job IDs in a GPU mapping file produce duplicate metric samples. The renderer rejects their identical labels, causing `/metrics` to return HTTP 500.

Deduplicate IDs within each file while preserving their first-seen order. The same job can still appear on multiple GPUs or MIG instances.

Add a regression using real mapping files and the metrics HTTP handler. It covers duplicate IDs, a job shared by two GPUs, and updated file contents on the next scrape. Document the duplicate-ID behavior.

Validation:
- The HTTP handler regression fails with status 500 on upstream and returns 200 with this change.
- Server and transformation package tests with `-race -short -count=1` passed.
- CPU short tests passed across 47 tested packages, excluding `internal/pkg/integration_test`.
- `make lint LINT_BASE_REV=HEAD`: 0 issues.
- `make check-fmt`: passed.

`make test-main` could not pass the DCGM integration package because `libdcgm.so.4` is unavailable in this environment. GPU/DCGM integration was not validated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ These mapping files follow a specific format:
 
 * Each file is named after either a unique GPU ID or a unique GPU ID and a GPU instance (MIG) ID separated with a "." (e.g., 0, 1, 2.0, 2.1, 3, etc.).
 * Each line in the file contains JOB IDs that run on the corresponding GPU/MIG instance.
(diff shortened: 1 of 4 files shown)


rvatkar:
LGTM.
cc: @nccurry

nccurry (approved):

Issue #731, opened 2026-09-08
Helm chart: default tolerations lack `nvidia.com/gpu`, so the DaemonSet schedules zero pods on tainted GPU nodes


KR-Ravindra:
### What is the version?

Helm chart 4.8.3 / dcgm-exporter 4.8.3 (`deployment/Chart.yaml` on `main` @ 181290c).

### What happened?

I installed the chart with `helm install --generate-name gpu-helm-charts/dcgm-exporter` on a cluster where the GPU node pool is tainted `nvidia.com/gpu=true:NoSchedule` (the usual pattern for Karpenter / Cluster Autoscaler GPU pools, EKS and GKE GPU node groups, and what the GPU Operator and device plugin expect). The release came up "successfully" but the DaemonSet had `DESIRED 0 / CURRENT 0`: no pods, no events, no metrics. Nothing in the chart output or README hints why.

The cause is the default tolerations in `deployment/values.yaml:190-193`:

```yaml
tolerations:
  - key: node-role.kubernetes.io/control-plane
    operator: Exists
    effect: NoSchedule
```

`deployment/templates/daemonset.yaml:74-77` renders exactly that list (`{{- with .Values.tolerations }}` / `toYaml`), so the DaemonSet controller marks every node carrying the `nvidia.com/gpu` taint ineligible. `helm template deployment/` confirms the rendered DaemonSet only tolerates the control-plane taint.

The NVIDIA device-plugin chart ships the missing toleration by default (`deployments/helm/nvidia-device-plugin/values.yaml:91-93`: key `nvidia.com/gpu`, `operator: Exists`, `effect: NoSchedule`), so installing both charts side by side gives you the device plugin on the GPU nodes and dcgm-exporter nowhere. The `dcgm-exporter.yaml` standalone manifest referenced in the README has no tolerations at all, same result.

### What did you expect to happen?

A GPU-only exporter DaemonSet should land on GPU nodes out of the box, the same way the device plugin does, or the install docs should say the taint has to be tolerated.

### What is the GPU model?

Not GPU-specific; the pods never get scheduled.

### What is the environment?

EKS (Kubernetes 1.35), GPU nodes tainted `nvidia.com/gpu=true:NoSchedule`, NVIDIA device plugin installed via its Helm chart with default values.

### How did you deploy the dcgm-exporter and what is the configuration?

Helm, default values:

```shell
[...]

PR #732, opened 2026-09-08
Tolerate nvidia.com/gpu taint in Helm chart and raw manifest defaults

KR-Ravindra:
## Problem

Installing the chart with default values on a cluster whose GPU nodes are tainted `nvidia.com/gpu=true:NoSchedule` (the pattern used by EKS/GKE GPU node groups, Karpenter/Cluster Autoscaler GPU pools, and what the NVIDIA device plugin and GPU Operator expect) produces a DaemonSet with `DESIRED 0 / CURRENT 0`. The release reports success, but no exporter pods run and no metrics are exposed. The `dcgm-exporter.yaml` standalone manifest has the same behaviour, and the README says nothing about taints.

## Root cause

`deployment/values.yaml:190-193` only ships a `node-role.kubernetes.io/control-plane` toleration:

```yaml
tolerations:
  - key: node-role.kubernetes.io/control-plane
    operator: Exists
    effect: NoSchedule
```

`deployment/templates/daemonset.yaml:74-77` renders that list verbatim (`{{- with .Values.tolerations }}` / `toYaml`), so nodes carrying the `nvidia.com/gpu` taint are ineligible for the DaemonSet. `dcgm-exporter.yaml` (DaemonSet spec, before `volumes:`) has no `tolerations` at all. The NVIDIA device plugin chart tolerates `nvidia.com/gpu` `Exists`/`NoSchedule` by default, so installing both side by side yields the device plugin on the GPU nodes and dcgm-exporter nowhere.

## Fix

- `deployment/values.yaml`: add an `nvidia.com/gpu` / `operator: Exists` / `effect: NoSchedule` toleration to the default list, matching the device plugin chart. Users who set `tolerations` themselves are unaffected because the template renders only what is in values.
- `dcgm-exporter.yaml`: add the same toleration to the standalone DaemonSet so the raw manifest stays aligned with the chart defaults.
- `README.md`: one sentence after the `helm install` step explaining which taints the defaults tolerate and that `tolerations` must be overridden for other taints.
- `tests/static/chart_scheduling_rendering_test.go` (new) plus a `tolerations` field on the existing render-contract structs in `tests/static/chart_rbac_rendering_test.go`: static chart tests asserting the default DaemonSet tolerates `nvidia.com/gpu`, that a user-supplied `tolerations` list replaces the defaults, and that `dcgm-exporter.yaml` carries the same toleration.

No chart version bump; versions are bumped in release commits in this repository.

## How tested

New static tests (`go test ./tests/static -run 'Toleration|Tolerates'`), run against the unchanged `values.yaml` / `dcgm-exporter.yaml` first:

```
--- FAIL: TestChartDefaultTolerationsRenderContract (0.01s)
    chart_scheduling_rendering_test.go:39: no Exists/NoSchedule toleration for taint "nvidia.com/gpu" in [{Key:node-role.kubernetes.io/control-plane Operator:Exists Effect:NoSchedule}]
--- FAIL: TestStandaloneManifestToleratesGPUTaint (0.00s)
    chart_scheduling_rendering_test.go:69: no Exists/NoSchedule toleration for taint "nvidia.com/gpu" in []
FAIL
FAIL	github.com/NVIDIA/dcgm-exporter/tests/static	0.056s
```

With the fix applied:

```
--- PASS: TestChartDefaultTolerationsRenderContract (0.01s)
--- PASS: TestChartTolerationsOverrideReplacesDefaults (0.01s)
--- PASS: TestStandaloneManifestToleratesGPUTaint (0.00s)
ok  	github.com/NVIDIA/dcgm-exporter/tests/static	0.040s
```

The full static suite (`make test-static` / `go test ./tests/static`) passes, and `gofmt -l tests/static` and `go vet ./tests/static` are clean.

Rendered the DaemonSet with default values before and after (`helm template test deployment/ --show-only templates/daemonset.yaml`).

Before (only the control-plane taint is tolerated; `grep -c nvidia.com/gpu` returns 0, exit 1):

```yaml
      tolerations:
      - effect: NoSchedule
        key: node-role.kubernetes.io/control-plane
        operator: Exists
      volumes:
```

After (`grep -c nvidia.com/gpu` returns 1, exit 0):

```yaml
      tolerations:
      - effect: NoSchedule
        key: node-role.kubernetes.io/control-plane
        operator: Exists
      - effect: NoSchedule
        key: nvidia.com/gpu
        operator: Exists
      volumes:
```

Also verified:

- `helm lint deployment/` passes (1 chart linted, 0 failed).
- Overriding the list still replaces the defaults entirely: `helm template test deployment/ --set 'tolerations[0].key=foo' --set 'tolerations[0].operator=Exists'` renders only the `foo` toleration.
- `dcgm-exporter.yaml` still parses as two documents (DaemonSet, Service) and the DaemonSet pod spec carries the new toleration.

Kubernetes e2e tests were not run (no GPU cluster available here).

## Links

- Issue: https://github.com/NVIDIA/dcgm-exporter/issues/731
- Device plugin default toleration: https://github.com/NVIDIA/k8s-device-plugin/blob/main/deployments/helm/nvidia-device-plugin/values.yaml

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,8 @@ helm install \
     gpu-helm-charts/dcgm-exporter
 ```
 
+The chart's default `tolerations` allow the DaemonSet to schedule on nodes tainted with `nvidia.com/gpu` (the taint used by the NVIDIA device plugin and GPU Operator) and on control-plane nodes. If your GPU nodes carry a different taint, override `tolerations` in your values, otherwise the DaemonSet will not schedule any pods on them.
+
 Once the `dcgm-exporter` pod is deployed, you can use port forwarding to obtain metrics quickly:
 
 ```shell
--- a/dcgm-exporter.yaml
+++ b/dcgm-exporter.yaml
@@ -70,6 +70,10 @@ spec:
           requests:
             cpu: 100m
             memory: 128Mi
+      tolerations:
+      - key: nvidia.com/gpu
+        operator: Exists
+        effect: NoSchedule
       volumes:
       - name: "pod-gpu-resources"
         hostPath:
--- a/deployment/values.yaml
+++ b/deployment/values.yaml
@@ -191,6 +191,9 @@ tolerations:
   - key: node-role.kubernetes.io/control-plane
     operator: Exists
     effect: NoSchedule
+  - key: nvidia.com/gpu
+    operator: Exists
+    effect: NoSchedule
 #- operator: Exists
 
 affinity: {}
--- a/tests/static/chart_rbac_rendering_test.go
+++ b/tests/static/chart_rbac_rendering_test.go
@@ -580,6 +580,7 @@ type chartResource struct {
 				AutomountServiceAccountToken *bool             `yaml:"automountServiceAccountToken"`
 				Containers                   []container       `yaml:"containers"`
 				ImagePullSecrets             []imagePullSecret `yaml:"imagePullSecrets"`
+				Tolerations                  []toleration      `yaml:"tolerations"`
 				Volumes                      []volume          `yaml:"volumes"`
 			} `yaml:"spec"`
 		} `yaml:"template"`
@@ -605,6 +606,13 @@ type container struct {
 
 type imagePullSecret struct {
 	Name string `yaml:"name"`
+}
+
+// toleration captures the pod toleration fields needed by the scheduling contract.
+type toleration struct {
+	Key      string `yaml:"key"`
+	Operator string `yaml:"operator"`
+	Effect   string `yaml:"effect"`
 }
 
 type serviceMonitorEndpoint struct {
--- a/tests/static/chart_scheduling_rendering_test.go
+++ b/tests/static/chart_scheduling_rendering_test.go
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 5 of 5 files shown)


KR-Ravindra (commented):
Round 1 self-review.

1. Checked `deployment/values.yaml` (only the control-plane toleration, L190-193) and `deployment/templates/daemonset.yaml` (`with .Values.tolerations` / `toYaml`, L74-77) on main; `dcgm-exporter.yaml` has no `tolerations` at all. The description matches the code.
2. Ran `go test ./tests/static -run 'Toleration|Tolerates'` on this branch: all three new tests pass, `gofmt -l tests/static` is clean. The helpers used (`renderChart`, `requireResourceKind`, `parseManifest`, `repoPath`) are the existing ones in `chart_rbac_rendering_test.go`; the only change there is the `Tolerations` field on the render-contract struct.
3. Overriding `tolerations` still replaces the list, so users with their own taints are unaffected; the test covers that case.

No Chart.yaml bump, consistent with this repo's release commits. Workflows need maintainer approval to run here; marking ready.

rvatkar:
cc: @nccurry

nccurry:
This is a breaking change. I think we can better document how to do this, but otherwise leave the defaults as is. 
Its easy enough to override helm defaults, which is what I would expect those downstream users to do.

KR-Ravindra:
@nccurry understood, keeping the defaults is fair. I will send a documentation-only follow-up: a short section in the chart README and the values.yaml comment showing the `nvidia.com/gpu` toleration override, so users who taint GPU nodes find it before the DaemonSet schedules zero pods. #731 can track that.

PR #733, opened 2026-09-09
docs(helm): document nvidia.com/gpu toleration override

KR-Ravindra:
## Problem

Installing the chart with default values on a cluster whose GPU nodes are tainted `nvidia.com/gpu=true:NoSchedule` (EKS/GKE GPU node groups, Karpenter/Cluster Autoscaler GPU pools, and the taint the NVIDIA device plugin chart tolerates by default) gives a DaemonSet with `DESIRED 0`: the release succeeds, no exporter pods run, and neither `deployment/values.yaml` nor the README mentions taints. #732 proposed adding the toleration to the defaults; the maintainer feedback there was that this is a breaking change and that the override should be documented instead, so this PR only documents it.

## Fix

- `deployment/values.yaml`: a comment block above `tolerations` explaining that a user-supplied list replaces the default rather than extending it, with an example that adds the `nvidia.com/gpu` toleration. The stray `#- operator: Exists` example line below the list is folded into the same comment. The default list itself is unchanged.
- `README.md`: a "Scheduling on tainted GPU nodes" subsection at the end of the Kubernetes quickstart with the same values snippet and one sentence on why the DaemonSet otherwise schedules no pods on those nodes.
- `deployment/README.md`: a "Scheduling on Tainted GPU Nodes" section under Configuration with the same snippet, so the chart's own README stays aligned with `values.yaml`.

No template, raw manifest, test, or default value changes, and no chart version bump.

## How tested

- `helm lint deployment/`: 1 chart linted, 0 failed.
- `helm template test deployment/` output is byte-identical before and after this change (only comments changed in `values.yaml`).
- `helm template test deployment/ -f values.yaml` with the documented override renders both the control-plane and `nvidia.com/gpu` tolerations on the DaemonSet, so the snippet does what the text says.

## Links

- Follow-up to https://github.com/NVIDIA/dcgm-exporter/pull/732
- Related to https://github.com/NVIDIA/dcgm-exporter/issues/731

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed it.


KR-Ravindra:
@nccurry this is the documentation-only follow-up from #732: defaults untouched, the README quickstart and the values.yaml comment now show the `nvidia.com/gpu` toleration override and note that a set list replaces the default. `helm template` output is byte-identical to main.

iacker:
The values.yaml comment reads well and the replace-not-merge behaviour matches the template.

One gap: deployment/README.md is the chart's own doc and has no tolerations section. deployment/AGENTS.md asks to keep values.yaml and README snippets aligned, so that file looks like the natural place for this too.

KR-Ravindra:
@iacker done in eed7817: added a "Scheduling on Tainted GPU Nodes" section to deployment/README.md (under Configuration) with the same override snippet, so the chart README, values.yaml comment and top-level README now say the same thing. helm lint passes and helm template output is still byte-identical to main. Thanks for catching it.

nccurry:
Set Helm `tolerations` for `nvidia.com/gpu` with `operator: Exists` and `effect: NoSchedule`. This is documented in 4.8.4. Closing.

PR #741, opened 2026-09-16
Remove the default CPU limit from chart resources

### LINKED ISSUE #740: Helm chart should not set a default CPU limit on dcgm-exporter

Change

Since 4.4.2-4.7.0 (commit 54267f2) the Helm chart ships concrete `resources` defaults:

```yaml
resources:
  limits:
    cpu: 200m
    memory: 512Mi
  requests:
    cpu: 100m
    memory: 128Mi
```

Two problems follow from this.

**1. The default CPU limit throttles the scrape path.** dcgm-exporter serves Prometheus scrapes. A `200m` CFS quota throttles it on a node with many GPUs, with a large metric list, or during a burst of concurrent scrapes — which surfaces as scrape timeouts and gaps in GPU telemetry exactly when the node is busy enough for that telemetry to matter. CPU limits on latency-sensitive system components are widely discouraged for this reason. The `requests` value already gives the scheduler what it needs and puts the pod in Burstable QoS; the limit adds throttling without adding protection.

**2. Non-empty defaults are not straightforwardly overridable.** Helm deep-merges maps, so a user who sets only `resources.limits.memory` silently inherits `limits.cpu: 200m` from the chart. Removing it requires `limits: {cpu: null}`, which is neither obvious nor documented. In practice the chart's opinion becomes a floor rather than a default.

Separately, the right memory number is fleet-dependent — it scales with GPUs per node, the size of the metric list and the scrape interval. The default already moved from `256Mi` to `512Mi` in 4.8.0 citing OOMs, which suggests a single number is a starting point rather than a recommendation, and it would help to say so in the file.

Finally, `deployment/values.yaml` and the raw `dcgm-exporter.yaml` manifest are currently out of sync: the manifest still carries `memory: 256Mi` and did not receive the 4.8.0 bump. `deployment/AGENTS.md` asks that values.yaml, templates and raw YAML examples stay aligned.

### Feature Description

As a cluster operator, I want the chart not to impose a CPU limit on dcgm-exporter by default, so that scrape collection is not throttled on busy GPU nodes and so that my own `resources` values are not partially overridden by chart defaults I did not ask for.

Keep the `requests` defaults and the memory limit, drop the default CPU limit, and add a short comment explaining the reasoning and the fact that the memory limit is a starting point. Bring `dcgm-exporter.yaml` back in line with the chart at the same time.

```yaml
resources:
  limits:
    memory: 512Mi
  requests:
    cpu: 100m
    memory: 128Mi
```

If you would rather keep a CPU limit, documenting the deep-merge behaviour in the chart README would at least make it discoverable that a partial `resources` override inherits the rest.

### Additional context

Happy to open a PR for this.

Noting that this is deliberately narrower than "revert to `resources: {}`": leaving a node-level DaemonSet with no requests at all would be its own problem (BestEffort, first evicted), so this keeps `requests` and targets only the part that causes harm.

ron-benaim-af:
Fixes #740.

Drops the default CPU limit introduced in 4.4.2-4.7.0 (54267f2) while keeping the requests and the memory limit.

dcgm-exporter serves the Prometheus scrape path. A 200m CFS quota throttles it on nodes with many GPUs, with a large metric list, or during concurrent scrapes — producing scrape timeouts and gaps in GPU telemetry exactly when the node is busy enough to care about. The `requests` value already gives the scheduler what it needs and keeps the pod in Burstable QoS.

This also keeps the chart from silently overriding user configuration: because Helm deep-merges maps, a user who sets only `resources.limits.memory` inherits `limits.cpu: 200m` unless they know to write `limits: {cpu: null}`.

While here, `dcgm-exporter.yaml` is realigned with the chart. It still carried `memory: 256Mi` and did not receive the 4.8.0 bump to 512Mi, which `deployment/AGENTS.md` asks be kept in sync.

No template changes — `resources` is already rendered conditionally, and the rendered output is unchanged apart from the absent CPU limit. `helm lint deployment/` passes and `helm template` renders the expected block.


nccurry:
Set `resources.limits.cpu: null` in your Helm values to remove the CPU limit. Closing.

